PR #97, opened 2015-01-05
Clarify use of public_updated_at field

tommyp:
The previous wording made it sound like the ContentStore used this field in some way. This new wording makes it clear that the publishing App should send the correct timestamp.

### DIFF
--- a/doc/content_item_fields.md
+++ b/doc/content_item_fields.md
@@ -114,9 +114,9 @@ for the Latin America and Caribbean region").
 
 ISO 8601 formatted timestamp. Present in all contexts.
 
-This is the time at which the content was last publicly visibly updated. This
-can be used for sorting by update date, but will not be changed for minor
-updates.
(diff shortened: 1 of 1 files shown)


doc/content_item_fields.md
```diff
@@ -115,8 +115,8 @@ for the Latin America and Caribbean region").
 ISO 8601 formatted timestamp. Present in all contexts.
 
 This is the time at which the content was last publicly visibly updated. This
-can be used for sorting by update date, but will not be changed for minor
-updates.
+is used for sorting documents by update date. It should change when there is a
+major update and should not change for a minor update.
```
alext: It might be better to word this differently.  Something like "This is the update date that should be surfaced to the user".  It's up to the publishing app when that should be.  It will typically coincide with a major update, but doesn't have to.

tommyp: Done.

PR #99, opened 2015-02-11
Require a base_path in the routes.

alext:
Using parens around a route parameter makes that param optional.  This
meant that requests for eg `/content` were being routed through to
`content_items#show`, which then errored because it expected `base_path` to be
populated.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -2,12 +2,12 @@ Rails.application.routes.draw do
 
   with_options :format => false do |r|
     r.with_options :constraints => {:base_path => %r[/.*]} do |path_routes|
-      path_routes.get "/content(*base_path)" => "content_items#show", :as => :content_item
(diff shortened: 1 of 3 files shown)


fatbusinessman:
Is this also going to fix the case when the request path is `/content/`? That produces the same error.

alext:
That shouldn't produce the error, and is actually a valid request (for the content item with a base_path of "/")  I wonder whether something else in the stack is normalising the request path and removing the trailing `/`

fatbusinessman:
> I wonder whether something else in the stack is normalising the request path and removing the trailing `/`

That sounds very plausible and familiar

alext:
Ok, I've done an experiment, and Rails is stripping the trailing `/` before it hits the router, so this currently won't support a base_path of `/`.

Having said that, I think that's a separate bug, and shouldn't block this PR.

PR #102, opened 2015-02-13
Don't generate rcov coverage data by default

jamiecobbett:
Only generate it on CI build, or if run with `COVERAGE=on`.

Generating the coverage files takes a non-trivial time after a successful run
of the test suite, which can be annoying if you want to get on with running
another command.


vinayvinay:
I've wanted this. Thanks!

PR #103, opened 2015-02-13
[DISCUSS] Validate that base_path in body matches the request URL

jamiecobbett:
We currently document base_path in the body as being required, but we don't
actually validate its presence or that it matches the one in the request URL.

This means we could end up with a situation where some publishing apps are
sending the wrong value some or all of the time but we don't realise it until
we (hypothetically) change content-store to start using the value in the body.

We still need to discard the base_path for the purposes of update_attributes,
as described here:
d70d92d98125e5dc44476ca9bd5051e677db59a2

The alternative to this change would be to drop `base_path` from the bodies for write requests.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -18,7 +18,7 @@ class ApplicationController < ActionController::API
   end
 
   def parse_json_request
-    @request_data = JSON.parse(request.body.read).except('base_path')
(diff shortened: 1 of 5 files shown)


jamiecobbett:
Closing in favour of  https://github.com/alphagov/content-store/pull/104

PR #104, opened 2015-02-16
Deprecate base_path in body in the publishing context

jamiecobbett:
The documentation implies that the base_path must be included in the request
body for write requests as well as in the URL. This is not true and in fact if
supplied in the body it is discarded, otherwise it is ignored - the base_path
in the URL is used instead.

Leading developers to believe that it is required but not performing any
validation on it seems likely to be a source of errors, so let's get rid of it.

Once all clients have been updated to omit it from the request body, we can
stop treating it as a special case and start rejecting it as we would any other
unexpected field.

### DIFF
--- a/doc/content_item_fields.md
+++ b/doc/content_item_fields.md
@@ -16,7 +16,8 @@ from the content store API can be found in
 
 ## `base_path`
 
-A string. Present in all contexts.
+A string. Present in the request URL in the storing context and included in the
+document in all other contexts.
 
 The absolute path of the content on GOV.UK. This uniquely identifies the
 content within the content store and allows the content store to answer the
--- a/doc/publish_intents.md
+++ b/doc/publish_intents.md
@@ -17,7 +17,8 @@ propagated out.
 
 A publish intent has the following fields:
 
-* `base_path` - The path of the item that will be published.
+* `base_path` - The path of the item that will be published. Found in the
+request URL of write requests and the response body for read requests.
 * `publish_time` - ISO 8601 formatted timestamp. The time the corresponding
   content item will be published.
 * `publishing_app` - The publishing app that owns this content (see
@@ -31,8 +32,8 @@ A publish intent has the following fields:
 For example:
 
 ``` js
+PUT /publish-intent/vat-rates
 {
-  "base_path": "/vat-rates",
   "publish_time": "2015-01-05T09:00:00+00:00",
   "publishing_app": "publisher",
   "rendering_app": "frontend",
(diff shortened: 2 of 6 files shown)


doc/content_item_fields.md
```diff
@@ -16,7 +16,8 @@ from the content store API can be found in
 
 ## `base_path`
 
-A string. Present in all contexts.
+A string. Present in the request URL in the publishing context, stored as `_id`
```
vinayvinay: i think you want to say: "Present in the request URL in the storing context, ..."
there are only 3 contexts we've established: storing, notifying, and retrieving.

vinayvinay: how about skipping "... stored as `_id`"? i feel that's extra information, because it's an implementation detail (mongo-specific). the next paragraph confirms that it is used as a unique identifier, which bit clients would be interested in knowing.

doc/output_examples/generic.json
```diff
@@ -1,5 +1,4 @@
 {
-  "base_path": "/path-on-gov-uk",
```
vinayvinay: possible oversight? we want the `base_path` in the retrieving context.

doc/content_item_fields.md
alext: :+1: I agree that the `_id` bit is an internal implementation detail that should never be surfaced to consumers of this API.  (The field is additionally [aliased as `base_path`](https://github.com/alphagov/content-store/blob/master/app/models/content_item.rb#L25) on the model, so even within the app, nothing should be referencing `_id` outside the model itself.)

doc/publish_intents.md
```diff
@@ -17,7 +17,6 @@ propagated out.
 
 A publish intent has the following fields:
 
-* `base_path` - The path of the item that will be published.
```
alext: I'd leave this here, but indicate that it's taken from the URL when writing, and included in the JSON when reading.

alext:
:+1: I think this is a good change.  I'm all for reducing duplication and potential confusion.

doc/publish_intents.md
jamiecobbett: Agreed - rebased.

doc/output_examples/generic.json
jamiecobbett: Oops, rebased.

doc/content_item_fields.md
jamiecobbett: Fair enough - rebased.

PR #106, opened 2015-02-20
Really remove non-case-study whitehall unpublishings

heathd:
In https://github.com/alphagov/content-store/pull/96 we attempted to
remove these.

However, we ran this migration before deploying the change to whitehall
which stopped sending unpublishings for non-case-studies to the content
store (deployed 22nd dec).

Therefore, some slipped through the net.

In addition we missed out unpublishings of translated content items.

This migration cleans up the remaining unpublishings.

### DIFF
--- a/db/migrate/20150220114658_really_remove_non_case_study_whitehall_unpublishings.rb
+++ b/db/migrate/20150220114658_really_remove_non_case_study_whitehall_unpublishings.rb
@@ -1,0 +1,42 @@
+class ReallyRemoveNonCaseStudyWhitehallUnpublishings < Mongoid::Migration
+  def self.up
+    puts "Removing redirects for #{content_items.count} items"
+    content_items.each do |content_item|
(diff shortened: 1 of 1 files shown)


db/migrate/20150220114658_really_remove_non_case_study_whitehall_unpublishings.rb
```diff
@@ -0,0 +1,45 @@
+class ReallyRemoveNonCaseStudyWhitehallUnpublishings < Mongoid::Migration
+  def self.up
+    puts "Removing redirects for #{content_items.count} items"
+    content_items.each do |ci|
+      puts ci.base_path
```
danielroseman: Why is this in a separate loop? Do you need it at all?

heathd: oops, I had put that in for debugging, will remove

PR #107, opened 2015-03-03
Dont return entity in PUT responses

jamiecobbett:
content-store returns the entity (the thing you sent it) as part of the response document, with an `errors` key mixed in. We don't have a use-case for this and we can't think of one, other than a stateless client (but we don't think this is a strong use case). If we need it, we can add it back in.

Aside:
This change makes some other work much simpler - https://github.com/alphagov/gds-api-adapters/pull/276

### DIFF
--- a/app/controllers/content_items_controller.rb
+++ b/app/controllers/content_items_controller.rb
@@ -59,6 +59,8 @@ class ContentItemsController < ApplicationController
     return_arbiter_error(:conflict, e)
   rescue GOVUK::Client::Errors::UnprocessableEntity => e
     return_arbiter_error(:unprocessable_entity, e)
+  rescue GOVUK::Client::Errors::BaseError => e
+    return_arbiter_error(:server_error, e)
   end
 
   def return_arbiter_error(status, exception)
--- a/app/controllers/content_items_controller.rb
+++ b/app/controllers/content_items_controller.rb
@@ -29,7 +29,9 @@ class ContentItemsController < ApplicationController
     else
       status = :unprocessable_entity
     end
-    render :json => PrivateContentItemPresenter.new(item), :status => status
+    response_body = {}
+    response_body[:errors] = item.errors.as_json if item.errors.any?
+    render :json => response_body, :status => status
   end
 
   private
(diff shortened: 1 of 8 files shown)


app/controllers/content_items_controller.rb
```diff
@@ -70,6 +70,6 @@ def return_arbiter_error(status, exception)
     else
       item.errors.add("url_arbiter_registration", "#{exception.response.code}: #{exception.response.raw_body}")
     end
-    render :json => PrivateContentItemPresenter.new(item), :status => status
+    render :json => { errors: item.errors.as_json }, :status => status
```
alext: Now that we're no longer returning the rest of the response JSON, there's no need to construct an unsaved ContentItem.  It would be simpler to just construct a hash with the error details in it.

jamiecobbett: Good spot - rebased.

jamiecobbett:
Rebased to add assertions about the response bodies after @heathd's comment on the now-orphaned commit.

spec/integration/publish_intent_crud_spec.rb
```diff
@@ -23,14 +23,11 @@
...
-      it "responds with a created status, and the intent as json" do
+      it "responds with a created status and an empty JSON document" do
         put_json "/publish-intent/vat-rates", data
 
         expect(response.status).to eq(201)
-
-        data = JSON.parse(response.body)
-        expect(data['base_path']).to eq('/vat-rates')
-        expect(data['publish_time']).to match_datetime(publish_time)
+        expect(response.body).to eq('{}')
```
vinayvinay: I'm thinking whether there's any value in returning an empty JSON document in the body?

For consumers of this API, it would be better to receive a `HEAD 201` with no body, instead of seeing a body that has no meaning. This also brings it closer to commonly followed API convention of returning a `HEAD 201` with a location header pointing to the newly created resource (which we don't need) in response to a PUT.

jamiecobbett: That's a good point. I did this to be consistent with the [proposed API guidelines](https://gov-uk.atlassian.net/wiki/display/PUB/RFC+6%3A+API+guidelines):
> Success AND error response bodies MUST be JSON, except where another format is desired (eg CSV)

What I haven't done in this PR is to follow this guideline as it felt like an unrelated change:
> Success AND error responses MUST include a top-level "message" key containing a user-friendly summary of the outcome - eg "Successfully created", "Validation failed",

heathd: given [Daniel's comment on the API guidelines RFC](https://gov-uk.atlassian.net/wiki/display/PUB/RFC+6%3A+API+guidelines?focusedCommentId=23855338#comment-23855338), I'll go ahead and merge this

PR #109, opened 2015-03-10
Remove url-arbiter integration

jamiecobbett:
publishing-api now does this for us, so we don't need to do it twice.

To be merged once https://github.gds/gds/puppet/pull/2631 has been deployed and we don't think we want to roll back.


vinayvinay:
:+1: Good to merge once you're ready.

PR #110, opened 2015-03-23
Allow the app to boot when rabbitmq is unavailable.

alext:
Change QueuePublisher to connect to rabbitmq on demand as opposed to in
the constructor. This means that the constructor won't raise an error
if rabbitmq is unavailable/unresponsive. Because this is called from an
initializer, if it raises an exception it will cause the application to
fail to boot.


bradleywright:
:+1:

PR #117, opened 2015-04-22
Add separate GET route for the public API to use

rboulton:
I'm shortly planning to make the content-store "GET content" API public.  As
part of that, we'd like to be very sure that only GET requests will be routed
from the public API.

To help with this, this commit adds a route to handle GET requests to
`/api/content/foo`.  This route serves exactly the same content as
`/content/foo`, but only handles GET requests.

By pointing publicapi at this route, we get some defence-in-depth that only the
read-only part of the API is available.  We also make it easier to distinguish
requests which have come via the public API from those which come from other
internal apps in log files and in systems such as Kibana.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -2,6 +2,9 @@ Rails.application.routes.draw do
 
   with_options :format => false do |r|
     r.with_options :constraints => {:base_path => %r[/.*]} do |path_routes|
+      # The /api/content route is used for requests via the public API
+      path_routes.get "/api/content*base_path" => "content_items#show", :as => :content_item_api
+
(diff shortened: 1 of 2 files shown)


alext:
This approach looks good to me. One slight issue - the api_urls for linked items returned in the JSON response will be incorrect as they will be the internal API url (eg 'https://content-store.production.alphagov.co.uk/content/vat-rates').  It would be good if these could return the external API url (eg `https://www.gov.uk/api/content/vat-rates`)

alext:
Summary of offline discussion:  we're going to pick up fixing the API urls in a follow-up PR.  We want to get this merged to unblock another piece of work.

PR #119, opened 2015-04-23
Return public API URLs when from the public endpoint

alext:
When requesting content items through the public API endpoint
(/api/content/foo), we want the API URLs returned in linked content
items to also use public API URLs.


rboulton:
I approve of this change, but I mostly paired on it with @alext, so we need 2nd eyes on it please.

jamiecobbett:
LGTM, but I'd appreciate someone who had more sleep reviewing too.

PR #121, opened 2015-04-30
Enable overriding of default_ttl from env var.

danielroseman:
This will allow the setting of a shorter timeout in the draft environments.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -42,8 +42,8 @@ module ContentStore
 
     # Caching
     config.cache_control_directive = 'public'
-    config.default_ttl = 30.minutes
-    config.minimum_ttl = 5.seconds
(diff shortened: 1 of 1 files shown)


config/application.rb
```diff
@@ -42,8 +42,8 @@ class Application < Rails::Application
 
     # Caching
     config.cache_control_directive = 'public'
-    config.default_ttl = 30.minutes
-    config.minimum_ttl = 5.seconds
+    config.default_ttl = ENV.fetch('DEFAULT_TTL', '1800').to_i
```
jamiecobbett: I think this could be: `ENV.fetch('DEFAULT_TTL', 30.minutes).to_i` and it would be more readable and comparable with the line below.

danielroseman: Done.

PR #130, opened 2015-06-26
Updating caching to use "Cache-Control" instead of "Expires"

tekin:
This updates content-store to set the "Cache-Control" max-age value instead of an "Expires" header, which is as good as deprecated in HTTP 1.1. This also allows us to remove some custom header manipulation code and use the built-in Rails helpers for setting page expiration.

Only the final commit makes the actual change. The first four commits re-organise the specs found in `fetching_content_item_spec.rb ` so that they are easier to understand and work with. `fetching_content_item_spec.rb` contained several nested levels of `describe`/`context` blocks which made it hard to reason about what was being tested and the set up that was active. Instead, I've pulled out the specs into separate files, grouped thematically on the particular feature under test.

### DIFF
--- a/spec/integration/api_request_spec.rb
+++ b/spec/integration/api_request_spec.rb
@@ -1,0 +1,14 @@
+require 'rails_helper'
+
+describe "API requests for content items", :type => :request do
+  let(:content_item) { create(:content_item, links: { 'related' => [linked_item.content_id] }) }
+  let(:linked_item) { create(:content_item, :with_content_id) }
+
+  it "corrrectly expands linked items with API URLs" do
+    get_api_content content_item
+
+    data = JSON.parse(response.body)
+
+    expect(data["links"]["related"].first["api_url"]).to eq("http://www.example.com/api/content#{linked_item.base_path}")
+  end
+end
--- a/spec/integration/fetching_content_item_spec.rb
+++ b/spec/integration/fetching_content_item_spec.rb
@@ -90,36 +90,6 @@ describe "Fetching a content item", :type => :request do
         end
       end
     end
-
-    describe "expanding linked items" do
-      # functional behaviour of link expansion covered in end_to_end_spec
-
-      describe "generating the API URL" do
-        let!(:linked_item) { create(:content_item, :with_content_id) }
-        before :each do
-          item.links["related"] = [linked_item.content_id]
-          item.save!
-        end
-
-        it "includes the API URL for linked items" do
-          get "/content#{item.base_path}"
-
-          expect(response.status).to eq(200)
-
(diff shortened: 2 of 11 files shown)


bradleywright:
:+1: for intent, @danielroseman or @alext would be good people to look at the code in detail.

spec/support/router_helpers.rb
```diff
@@ -3,6 +3,14 @@
 module RouterHelpers
   include GdsApi::TestHelpers::Router
 
+  def get_content(content_item)
+    get "/content#{content_item.base_path}"
+  end
+
+  def get_api_content(content_item)
+    get "/api/content#{content_item.base_path}"
+  end
```
alext: I don't think these belong in this helper.  It's all related to the router, and registering routes there, whereas these helpers are related to making requests to the content-store.

spec/integration/fetching_content_item_spec.rb
```diff
@@ -1,217 +1,59 @@
...
-        end
-
-        it "includes the public API URL when requested through the public endpoint" do
-          get "/api/content#{item.base_path}"
-
-          expect(response.status).to eq(200)
-
-          data = JSON.parse(response.body)
-          expect(data["links"]["related"].first["api_url"]).to eq("http://www.example.com/api/content#{linked_item.base_path}")
-        end
```
alext: I don't see these 2 tests covered in the restructured tests.  These specifically cover the detail where the `api_url` is different depending which route you request the item through (ie with and without the `/api` prefix)

spec/integration/api_request_spec.rb
```diff
@@ -0,0 +1,14 @@
+require 'rails_helper'
+
+describe "API requests for content items", :type => :request do
```
alext: I'm not quite clear on the scope of this test.  Aren't all the GET requests API requests for content items?

alext:
Code changes look good.  I have a few comments on the restructuring of the tests.

spec/integration/api_request_spec.rb
tekin: Ah, maybe this would be better named `public_api_request_spec.rb`, i.e. it's for specs covering the public-facing `/api` endpoint. The one spec within replaces one the two tests mentioned in the other comment above (`"includes the public API URL when requested through the public endpoint"`). The other one was lost in translation, I'll re-instate it somewhere.

spec/support/router_helpers.rb
tekin: Good point. I've moved them to `RequestHelpers`.

tekin:
@alext thanks for the comments, I think I've addressed them all.

PR #131, opened 2015-06-30
Don't send access_limited items on queue.

danielroseman:
As items that are explicitly access limited are only intended for viewing by the listed individuals, and we aren't currently enforcing access limiting anywhere else, they should not be sent on the message queue.

### DIFF
--- a/spec/models/content_item_spec.rb
+++ b/spec/models/content_item_spec.rb
@@ -211,6 +211,20 @@ describe ContentItem, :type => :model do
       end
     end
 
+    context "queue publishing" do
+      it "sends the item on the message queue" do
+        expect(Rails.application.queue_publisher).to receive(:send_message)
+        item = build(:content_item)
+        item.upsert
+      end
+
+      it "does not send the item on the message queue for access_limited items" do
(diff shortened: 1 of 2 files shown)


mattbostock:
Could you please include the detail from your PR description in your commit message?

danielroseman:
@mattbostock done.

mattbostock:
Build failure :crying_cat_face:

danielroseman:
Hmm, re-running the build worked - seems like an existing test is a bit too sensitive to timing issues. I'll fix in a separate PR.

mattbostock:
:+1:

I don't know enough about what the queue does to know if this is safe to merge. I'd be happy for either that info to be added to the commit or for someone else with better context to merge.

spec/models/content_item_spec.rb
```diff
@@ -211,6 +211,20 @@
       end
     end
 
+    context "queue publishing" do
+      it "calls send_message on upsert" do
```
heathd: rather than 

```ruby
it "calls send_message on upsert"
```
how about 

```ruby
it "sends the item on the message queue"
```

danielroseman: Done.

PR #133, opened 2015-06-30
Make cache test less sensitive to time.

danielroseman:
Occasionally the test suite would fail on an expectation that the cache
timeout was exactly 300 seconds from now. Use Timecop.freeze to ensure
that no system time elapses between creating the item and verifying its
timeout.

### DIFF
--- a/spec/integration/fetching_publish_intent_spec.rb
+++ b/spec/integration/fetching_publish_intent_spec.rb
@@ -68,6 +68,7 @@ describe "Fetching a content item with a publish intent", :type => :request do
 
   context "a publish intent before the default TTL time" do
     before(:each) do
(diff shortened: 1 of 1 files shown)


spec/integration/fetching_publish_intent_spec.rb
```diff
@@ -89,6 +89,7 @@
 
 describe "Fetching a publish intent without a content item", :type => :request do
   before(:each) do
+    Timecop.freeze
```
alext: Do you need the same thing on the block above (line 69)?

danielroseman: Good call, added.

PR #138, opened 2015-08-12
Add phase to Content Item

tommyp:
In order to mark an item as either Alpha or Beta a `phase` attribute
has been added. This should only be `alpha`, `beta` or `nil` and can be
used on the Frontend to alter functionality or visual elements based on
its' state.

### DIFF
--- a/app/models/content_item.rb
+++ b/app/models/content_item.rb
@@ -37,6 +37,7 @@ class ContentItem
   field :redirects, :type => Array, :default => []
   field :links, :type => Hash, :default => {}
   field :access_limited, :type => Hash, :default => {}
+  field :phase, :type => String
   attr_accessor :update_type
 
   scope :renderable_content, -> { where(:format.nin => NON_RENDERABLE_FORMATS) }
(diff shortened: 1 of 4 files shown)


app/models/content_item.rb
```diff
@@ -52,6 +53,10 @@ def self.create_or_replace(base_path, attributes)
   validate :no_extra_route_keys
   validate :links_are_valid
   validate :access_limited_is_valid
+  validates :phase,
+            inclusion: { in: ['alpha', 'beta'],
+                         allow_blank: true,
```
alext: We should probably use `allow_nil` here, otherwise this will also accept an empty string.

spec/models/content_item_spec.rb
```diff
@@ -614,5 +614,34 @@
         end
       end
     end
+
+    describe 'phase' do
```
alext: This block doesn't really belong here (it's in the "linked items" describe block").  It would be better in the "validations" block at the top of the file.

tommyp:
@alext Sorted both of your comments.

spec/models/content_item_spec.rb
```diff
@@ -263,6 +263,35 @@
...
+      end
+
+      context 'a content item with a phase other than alpha or beta' do
+        let(:content_item) { build(:content_item, phase: 'not-a-correct-phase') }
+
+        it 'is invalid' do
+          expect(content_item).to_not be_valid
+        end
+      end
+    end
```
alext: Sorry to nitpick, but this would probably be best up with the other "regular" fields (maybe after the `content_id` validations).  It would also be good if it followed the pattern of the other validations - it makes it easier to read if all the tests in a given file follow a similar pattern.

tommyp: Moved it further up.

PR #139, opened 2015-08-12
Add analytics identifier to links hash if exists

edds:
If the analytics identifier exists in the details hash then it should be
part of the presented links hash object. This will let us add analytics
tracking for content which has an organisations associated to it.

https://trello.com/c/uXiRRmb1/74-send-organisation-data-to-google-analytics-for-pages-in-government-frontend-medium

### DIFF
--- a/app/models/content_item.rb
+++ b/app/models/content_item.rb
@@ -116,7 +116,7 @@ private
       .renderable_content
       .where(:content_id => {"$in" => links.values.flatten.uniq})
       .where(:locale => {"$in" => [I18n.default_locale.to_s, self.locale].uniq})
-      .only(:content_id, :locale, :base_path, :title, :description)
+      .only(:content_id, :locale, :base_path, :title, :description, :details)
       .sort(:updated_at => -1)
       .group_by(&:content_id)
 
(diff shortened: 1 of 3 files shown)


alext:
I'm not convinced this change is good for a number of reasons.

* There's a performance penalty to loading the details hash for every linked item.
* the linked item presentation is no longer a subset of the public presentation (the field is no longer nested in the details hash).

I wonder whether it would be better for government-frontend to make a second call to the content-store for the organisation details, rather than imposing this penalty on every linked item for every request to the content-store.

jamiecobbett:
@alext some content items are linked to multiple organisations.

edds:
I am very happy to discuss approach and come up with a better way of doing this if there is one. 

This will effect more than just government frontend though. Every content item that can be tagged to an organisation, worldwide org or worldwide location should be exposing this information. Finder frontend is the other application where there is a pressing need to surface this information (for policy pages). But there are many many more types of things which aren't just that application which will need this information (campaign pages and specialist pages are other examples).

jamiecobbett:
We could easily move the `analytics_identifier` to the top level of the content item and make it optional - nothing is using it yet. That should alleviate @alext's concerns.

edds:
Closing in favour of https://github.com/alphagov/content-store/pull/143 and https://github.com/alphagov/govuk-content-schemas/pull/109

PR #140, opened 2015-08-13
Only register routes for an item when they've changed.

alext:
In order to avoid unnecessary router reloads, we should avoid sending
routes to the router when they haven't changed.

This updates the router reload logic to take note of the previously
existing content_item, and skip registering routes, if there is a
previous item with an identical route set.

This will make most bulk re-registering processes much more efficient.


jamiecobbett:
:+1:

bradleywright:
:+1:

PR #141, opened 2015-08-13
Accept root path

elliotcm:
Also updates the routing to be compatible with
newer Rails versions.

Needed for https://trello.com/c/blLdEZN5/292-make-apps-register-special-routes-on-deploy

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -21,6 +21,10 @@ class ApplicationController < ActionController::API
   end
 
   def encoded_base_path
-    URI.escape(params[:base_path])
+    URI.escape(base_path)
+  end
+
+  def base_path
+    "/#{params[:base_path_without_root]}"
   end
 end
--- a/app/presenters/content_item_presenter.rb
+++ b/app/presenters/content_item_presenter.rb
@@ -49,7 +49,7 @@ private
   end
 
   def api_url(item)
-    @api_url_method.call(item.base_path)
+    @api_url_method.call(item.base_path_without_root)
   end
 
   def web_url(item)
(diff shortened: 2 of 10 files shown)


app/controllers/application_controller.rb
```diff
@@ -21,6 +21,10 @@ def parse_json_request
   end
 
   def encoded_base_path
-    URI.escape(params[:base_path])
+    URI.escape(base_path)
+  end
+
+  def base_path
+    "/#{params[:base_path]}"
```
alext: I do wonder whether this param name is still right. It feels that it could be confusing that `params[:base_path]` doesn't actually contain a base_path.  Maybe we should rename it to something like `base_path_without_slash` to make it very clear.

heathd: I'd be in favour of that

app/presenters/content_item_presenter.rb
```diff
@@ -37,7 +37,7 @@ def present_linked_item(linked_item)
   end
 
   def api_url(item)
-    @api_url_method.call(item.base_path)
+    @api_url_method.call(item.base_path.sub(%r{^/}, ""))
```
heathd: the need to strip the leading slash is coupled to the implementations of the `content_item_api_url` and `content_item_url` methods which are a few hops away in the code. 

How about refactoring this so that `@api_url_method.call` accepts an `item` and does what it needs. Or even continues to accept `item.base_path` and does the stripping itself. That moves the knowledge about the url making into one place.

app/controllers/application_controller.rb
elliotcm: Updated.

heathd:
:+1:

PR #143, opened 2015-08-13
Add analytics identifier to root object

edds:
Analytics identifiers are short codes which we can use to represent
objects in character restricted environments (like Google Analytics
custom dimensions). Having them on the root lets us easily pull them out
to be part of the links hash so that content can be correctly marked
with all related dimensions.

### DIFF
--- a/spec/factories/content_item.rb
+++ b/spec/factories/content_item.rb
@@ -8,7 +8,6 @@ FactoryGirl.define do
     publishing_app 'publisher'
     update_type 'minor'
     routes { [{ 'path' => base_path, 'type' => 'exact' }] }
-    phase nil
 
     trait :with_content_id do
       content_id { SecureRandom.uuid }
--- a/spec/models/content_item_spec.rb
+++ b/spec/models/content_item_spec.rb
@@ -106,6 +106,17 @@ describe ContentItem, :type => :model do
       end
     end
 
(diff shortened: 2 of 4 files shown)


edds:
Needed for https://github.com/alphagov/govuk-content-schemas/pull/109

spec/models/content_item_spec.rb
```diff
@@ -106,6 +106,17 @@
       end
     end
 
+    context 'analytics_identifier' do
+      it 'allows no analytics_identifier' do
```
jamiecobbett: `is valid without an analytics_identifier` might read better?

spec/factories/content_item.rb
```diff
@@ -9,6 +9,7 @@
     update_type 'minor'
     routes { [{ 'path' => base_path, 'type' => 'exact' }] }
     phase nil
+    analytics_identifier nil
```
alext: Do we need to add this to the factories?  Won't it default to `nil` anyway?  (I don't think it was needed for phase either...)

spec/models/content_item_spec.rb
edds: good idea, updated.

spec/factories/content_item.rb
edds: removed both. Tests still pass.

PR #145, opened 2015-08-18
Add analytics identifiers to links hash objects

edds:
So that we can correctly set any related analytics identifiers on
frontend applications expose the analytics identifiers for all the items
in the links hash if they exist.

This is the new approach since the discussions which resulted from here: https://github.com/alphagov/content-store/pull/139

https://trello.com/c/uXiRRmb1/74-send-organisation-data-to-google-analytics-for-pages-in-government-frontend-medium


jamiecobbett:
I'll give others a chance to comment, but :+1:

PR #146, opened 2015-09-09
Allow phase to be live, and require a phase

alext:
This also makes phase default to 'live' when not given because that's generally correct.

This PR is the implementation of alphagov/govuk-content-schemas#114


alext:
Added a migration to populate the phase for existing items.

PR #147, opened 2015-09-14
Include content_id in frontend representations.

alext:
As described in alphagov/govuk-content-schemas#121, it's becoming
nexessary for some of the frontend applications to know about
content_ids. This exposes them in both the base item, and the expanded
links.

This includes a change to stop returning self in available translations for
an item with no content_id (see commit c89f3c2 for details)

### DIFF
--- a/app/models/content_item.rb
+++ b/app/models/content_item.rb
@@ -160,7 +160,7 @@ private
     ContentItem
       .renderable_content
       .where(:content_id => content_id)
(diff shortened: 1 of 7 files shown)


rboulton:
Content-schemas PR is now merged.

rboulton:
This looks good to me.  I'm not 100% certain that the change to the representation of the available_translations is correct (and that nothing depends on it) though, so I'd like someone who's worked on that to review this.

PR #150, opened 2015-09-18
Add a rake task that reports on duplicate content IDs

mobaig:
Why?:

We want to eventually use content IDs to uniquely identify content in
the publishing API. This report is a starting point for assessing the
different cases where content IDs are non-unique.

How?:

The rake task does a simple count of duplicate content ID occurrences
using a mongo aggregation. It drops a list of content items into a CSV
(displaying a 'useful' subset of a content item's fields) and writes this
to $PROJECT_ROOT/tmp.

Related PR: https://trello.com/c/uivlG9td


tijmenb:
Looks good to me. :+1: 

Would nesting this in the `data_hygiene` namespace make sense? (https://github.com/alphagov/content-store/blob/master/lib/tasks/data_hygiene.rake)

mobaig:
@tijmenb added a commit that shuffles the hygiene tasks around. WDYT?

bishboria:
:+1:

tijmenb:
@mobaig Yes, perfect. I really like the directory-per-namespace, it makes searching for the task much easier.

PR #153, opened 2015-10-13
Fix inconsistent redirects for policies

tijmenb:
In New World Publishing, applications send content to the publishing-api, which forwards it to the content-store and also registers routes for the document with the router-api.

However, in the past apps have sent routes directly to router-api. This has happened frequently when setting up redirects.

It means we have a bunch of routes in the router for which either no content items exist, or for which the content item doesn't match the content item. 

The rake task added in the first commit is used to find a subset of the last case. It finds policies with format "placeholder" in the content-store that are actually redirects. The outputted information is used to bring the database up-to-date by turning the content items into [proper redirect content items](https://github.com/alphagov/content-store/blob/master/doc/redirect_item.md).

I've extracted the `InconsistentRedirectFinder` to enable us to re-use the code to fix other inconsistent redirects. The `rake data_hygiene:inconstent_policy_redirects` task is temporary and will be removed once we've fixed them.

The motive to fix the policies now is that they have duplicate content_ids and we need to fix that for tagging.

Trello: https://trello.com/c/fhz6hWaA/367-fix-duplicate-content-id-for-policies


lib/tasks/data_hygiene/inconsistent_redirect_finder.rb
```diff
@@ -0,0 +1,16 @@
...
+  def initialize(content_items)
+    @content_items = content_items
+  end
+
+  def items_with_inconsistent_redirects
+    content_items.select do |content_item|
+      next if content_item.format == "redirect"
+
+      route = Rails.application.router_api.get_route(content_item.base_path)
+      route && route.handler == "redirect" && !route.redirect_to.to_s.include?('#')
```
alext: Why the special handling for redirect destinations that include a `#`?

rboulton: It may just be me, but I'm having trouble understanding this. In particular, what's the check with the '#' trying to do?

tijmenb: I put that in to filter the policies that redirected to anchor tags:

```
/government/policies/helping-developing-countries-economies-to-grow/supporting-pages/the-investment-facility-for-utilising-uk-specialist-expertise-ifuse
```

redirects to:

```
/government/publications/2010-to-2015-government-policy-economic-growth-in-developing-countries/2010-to-2015-government-policy-economic-growth-in-developing-countries#appendix-7-the-investment-facility-for-utilising-uk-specialist-expertise-ifuse
```

But I guess that should have gone into the rake task to find the policies as it's specific to that use case. Removed in https://github.com/alphagov/content-store/commit/2160bfb0e3931e1525b105552311d9fe9d196167.

tijmenb: @rboulton does it make sense now?

tijmenb:
Since I've also added https://github.com/alphagov/content-store/commit/afe84fa60f6404dded1331bac9be8a2d5e3e529d, I've updated PR title + description.

rboulton:
This makes more sense - thanks.

PR #155, opened 2015-10-14
Add rake task for exporting content items to JSON.

elliotcm:
https://trello.com/c/AgfGK1lz/325-importing-existing-data

This produces a file with one JSON document per line
in ./tmp with each document having the structure:

```
{
  updated_at: "2015-10-14T15:07:09+01:00",
  content_item: {
    content_id: "2e2d3551-4147-4136-8c0d-f0904b69c8c6",
    etc..
  }
}
```

These files are intended for use by the publishing
API when importing documents.

Example usage:
```
$ rake data_hygiene:export_content_items:all
Progress [====>................................................................................................] (4%)
```

### DIFF
--- a/lib/tasks/data_hygiene/export_data.rb
+++ b/lib/tasks/data_hygiene/export_data.rb
@@ -1,0 +1,44 @@
+module Tasks
+  module DataHygiene
+    class ExportData
+      def initialize(file, stdout)
+        @file = file
+        @stdout = stdout
+      end
+
+      def export_all
+        total = ContentItem.count
+
+        ContentItem.all.each.with_index(1) do |content_item, index|
+          content_item_hash = content_item.as_json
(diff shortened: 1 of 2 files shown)


lib/tasks/data_hygiene/export_data.rb
```diff
@@ -0,0 +1,44 @@
+module Tasks
+  module DataHygiene
+    class ExportData
+      def initialize(file, stdout)
```
mobaig: any reason we're passing stdout as a parameter here? what's the intended purpose? looking at usage below, we could do without and just use naked calls to `puts` and `print`.

elliotcm: I wasn't keen to have a worker object bypassing its wrapper and talking directly to the outside world.  This allows non-error output to be sent to a file within the rake task if we chose to.

PR #156, opened 2015-10-15
Add redirects for statistics-publications

tijmenb:
The publications in this CSV are being redirected by Whitehall from `/government/publications/:id` to `/government/statistics/:id`. Adding a redirect item for these fixes the problem that they have the same content_id.


tijmenb:
Amended to use the same filename for CSV as the migration. @alext

PR #157, opened 2015-10-15
Add redirects for school commissioner URLs and like pages

tijmenb:
These URLs redirect to a new spelling of the organisation, much like
the RAIB case (#154). In this case we can't find redirects in
router-data, so we assume they have been added via another way.

We need to make them redirects without content_id because the
content_id are duplicated.


tijmenb:
Amended to use the same filename for CSV as the migration. @alext

tijmenb:
Added a commit to fix a very similar case I found.

tijmenb:
Added another commit to replace what we've removed here: https://github.gds/gds/router-data/pull/487

PR #158, opened 2015-10-15
Remove content items that 404

tijmenb:
The paths in the CSV return 404 on the site. For example:

https://www.gov.uk/government/collections/ots-tax-complexity

However, they do have content items.

https://www.gov.uk/api/content/government/collections/ots-tax-complexity

The content_ids for these items are duplicated. In the above case it duplicates:

https://www.gov.uk/api/content/government/collections/hmt-ots-tax-complexity

Because they return 404, we think its best to remove these content items, as the items have not previously worked.


alext:
Look good.

One thought though: In other apps, we have a convention where a CSV file needed by a migration is given the same name as the migration and kept alongside it.  This makes it easy to see where the various CSV files are used. Probably worth keeping that convention here.

tijmenb:
Good point, amended.

PR #159, opened 2015-10-15
Add redirects for previously 404'ing pages

tijmenb:
These two pages return 404, but have the same content_id as similar pages.

- /government/world/organisations/british-trade-cultural-office-taiwan/about/recruitment is a sub page of an organisation that has been renamed in https://github.com/alphagov/whitehall/pull/2245. This subpage have not been renamed in this case. This PR fixes that.

- /government/organisations/court-of-protection seems to have existed but dissapeared. It has the same content_id as /government/organisations/court-of-protection

Trello: https://trello.com/c/cme0kKzQ/366-fix-duplicate-content-ids


bishboria:
:+1:

PR #160, opened 2015-10-16
Remove knowledge of update type.

elliotcm:
This was previously used in the representation which
was sent to the message bus.  This sending is now
done in the Publishing API, so Content Store no
longer uses it.

Because this was a virtual field on the content
item the field was never stored in the database.

https://trello.com/c/0vlt6iS0/355-remove-mentions-of-update-type-from-content-store

### DIFF
--- a/doc/content_item_fields.md
+++ b/doc/content_item_fields.md
@@ -239,22 +239,6 @@ array contains an original path, a routing type, and a destination path. See
 For redirect content items, the behaviour is slightly different.  See
 redirect_item.md for details.
 
-## `update_type`
-
-A string. Present in storing and notifying contexts.
-
-This indicates the type of update that was made to the content item.
-It must be one of:
-
- - 'major' - major changes to a piece of content.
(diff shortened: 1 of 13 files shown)


doc/content_item_fields.md
```diff
@@ -239,22 +239,6 @@ array contains an original path, a routing type, and a destination path. See
...
-
- - 'major' - major changes to a piece of content.
- - 'minor' - changes which don't affect the meaning of the
-   content, eg typo correction.
- - 'republish' - useful in situations such as when the data
-   structure has changed.
-
-Other types may be added in future, the content store will just pass them through
-to the fanout.
-
```
alext: We need to think about this removal.  At the moment, this is the only documentation for what gets sent to the publishing-api, and this fields is still required there.

elliotcm: It's also out of date though.  We need to write a bunch of docs for the new publishing API.

PR #162, opened 2015-10-29
Add data migration to include parent content ID on email signups

mobaig:
This is intended to be used by the email-alert-frontend to send on to the
email-alert-api, allowing policy subscriptions to be created by content
ID.

Trello: https://trello.com/c/ZD65QUBN/326-email-alerts-based-on-links-hash

### DIFF
--- a/db/migrate/20151029154403_add_content_ids_to_email_signups.rb
+++ b/db/migrate/20151029154403_add_content_ids_to_email_signups.rb
@@ -1,0 +1,17 @@
+class AddContentIdsToEmailSignups < Mongoid::Migration
+  def self.up
+    ContentItem.any_of({base_path: /government\/policies\/.*email-signup$/}).each do |content_item|
+      policy_slug = content_item.details["tags"]["policy"].first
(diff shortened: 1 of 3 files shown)


db/migrate/20151029154403_add_content_ids_to_email_signups.rb
```diff
@@ -0,0 +1,16 @@
...
+      ).first.content_id
+      content_item.links[:parent] = [policy_content_id]
+      content_item.save!
+    end
+  end
+
+  def self.down
+    # noop
+  end
+end
```
jamiecobbett: Nitpick: valid *nix files have a trailing newline.

mobaig: amended.

mobaig:
Closing this temporarily while I rethink the order of these policy-related changes - I suspect policy publisher should be updated first so that email-signups carry the links data by default.

mobaig:
policy-publisher has been updated as required: https://github.com/alphagov/policy-publisher/pull/111

PR #163, opened 2015-10-30
De-duplicate content items...

steventux:
Duplicate content items sharing the same `content_id` and `locale`
periodically get added to the content store. A detailed explanation
of how this happens can be found here:
https://gov-uk.atlassian.net/wiki/display/FS/Duplicate+content-ids

This pull request adds tasks to find duplicates and report or
de-duplicate by destroying the older record(s) in a duplicate set.

To report:
```
rake data_hygiene:content_items:report_duplicates
```

To deduplicate:
```
rake data_hygiene:content_items:deduplicate
```

### DIFF
--- a/spec/lib/tasks/data_hygiene/content_item_deduplicator_spec.rb
+++ b/spec/lib/tasks/data_hygiene/content_item_deduplicator_spec.rb
@@ -1,0 +1,44 @@
+require "rails_helper"
+require "tasks/data_hygiene/content_item_deduplicator"
+
+describe Tasks::DataHygiene::ContentItemDeduplicator do
+  let(:fake_stdout) { StringIO.new }
+
+  before do
+    @real_stdout = $stdout
+    $stdout = fake_stdout
+  end
+
+  after { $stdout = @real_stdout }
+
+  describe "#deduplicate" do
+    it "runs without issue" do
+      content_item = create(:content_item_with_content_id, locale: 'cy')
+      dupe = create(:content_item, content_id: content_item.content_id, locale: 'cy')
+      locale_variant = create(:content_item, content_id: content_item.content_id, locale: 'en')
+
+      expect { subject.deduplicate }.to change(ContentItem, :count).by(-1)
+
+      fake_stdout.rewind
+      output = fake_stdout.read
+      expect(output).to match(/2 duplicates found/)
+      expect(output).to match(/1 records were removed/)
+    end
+  end
+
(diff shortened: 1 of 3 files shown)


jackscotti:
This looks good but I think there should be tests around `ContentItemDeduplicator`.

steventux:
Thanks @jackscotti, updated.

spec/lib/tasks/data_hygiene/content_item_deduplicator_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+    it "runs without issue" do
+      content_item = create(:content_item_with_content_id, locale: 'cy')
+      dupe = create(:content_item, content_id: content_item.content_id, locale: 'cy')
+      locale_variant = create(:content_item, content_id: content_item.content_id, locale: 'en')
+
+      expect { subject.deduplicate }.to change(ContentItem, :count).by(-1)
+
+      fake_stdout.rewind
+      output = fake_stdout.read
+      expect(output).to match(/2 duplicates found/)
```
benlovell: There are output matchers that let you assert on `stdout` and `stderr`. You could go fully end-to-end and also remove the need to overwrite them in the tests. Something like:

```ruby
expect { subject.deduplicate }.to output('2 duplicates found').to_stdout
# etc
```

spec/lib/tasks/data_hygiene/content_item_deduplicator_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+      content_item = create(:content_item_with_content_id, locale: 'en')
+      create(:content_item, content_id: content_item.content_id, locale: 'en')
+      create(:content_item, content_id: content_item.content_id, locale: 'fr')
+
+      expect { subject.report_duplicates }.not_to change(ContentItem, :count)
+
+      fake_stdout.rewind
+      output = fake_stdout.read
+
+      expect(output).to match(/2 duplicates found/)
```
benlovell: Same here of course.

spec/lib/tasks/data_hygiene/content_item_deduplicator_spec.rb
steventux: I like the simplicity of this output matcher but my primary concern (at least for the deduplicate test) is to assert a record has been destroyed, is there a neat way to place multiple matchers around the ```expect { subject.deduplicate }.to ...` ?

PR #166, opened 2015-11-09
Protect against out of order messages

tuzz:
This work makes it so that the publishing api sends a transmitted_at timestamp (in nanoseconds) to the content store for content item /v1 and /v2 incoming requests. We discussed this with @danielroseman and @bradwright and we think this is the best approach for now (without worrying too much about content item vs. link set race conditions). If we find that race conditions are an issue (we don't think they will be), we've got an upgrade path in mind for the future.

This work also sets up a pact for this specific piece of work between the publishing api and the content store. There's probably a job to back fill the pacts for the other requests that are made between these applications at some point in the future. As part of this set up work, we push the pact json file up to the pact broker. When we spoke to Alex he said that this was the right thing to do. We've more-or-less copied the rake task implementations for this from gds-adapters/publishing-api.

**Note:** The content store pull request will not pass on Jenkins until the Publishing API pull request in merged.

The sister pull request in the publishing api is here: https://github.com/alphagov/publishing-api/pull/127

https://trello.com/c/eQPSFnwP/370-deal-with-out-of-order-updates-in-content-store


app/controllers/content_items_controller.rb
```diff
@@ -20,14 +20,21 @@ def update
...
-    if result
-      status = (result == :created ? :created : :ok)
-    else
+    response_body = {}
+    case result
+    when :created
+      status = :created
+    when :stale
+      status = :conflict
+      response_body = { errors: "A later transmitted_at is already stored" }
```
elliotcm: Could we have a more useful error message than this?

app/models/update_lock.rb
```diff
@@ -0,0 +1,25 @@
+class UpdateLock
+  attr_reader :lockable, :locked_at
```
elliotcm: These can be private.

app/controllers/content_items_controller.rb
elliotcm: Also, this message will be returned if `transmitted_at` isn't implemented, discarding the error message from `UpdateLock`.

app/models/update_lock.rb
```diff
@@ -0,0 +1,25 @@
...
+    @lockable = lockable
+
+    if lockable && !lockable.respond_to?(:transmitted_at)
+      raise OutOfOrderTransmissionError.new("#{lockable.class} must implement transmitted_at")
+    end
+  end
+
+  def check_availability!(transmitted_at)
+    raise MissingAttributeError, "transmitted_at is mandatory" unless transmitted_at
+    transmitted_at = Float(transmitted_at)
```
elliotcm: I'm real worried about using floats across two systems (via JSON) when we're using them specifically for the tiny fractions of a second.

elliotcm: That pun wasn't even intentional.

app/models/update_lock.rb
```diff
@@ -0,0 +1,25 @@
...
+
+    if lockable.present? && lockable.transmitted_at >= transmitted_at
+      error = "Tried to process request with transmission time #{transmitted_at},"
+      error += " but the latest #{lockable.class} has a newer (or equal) timestamp of #{lockable.transmitted_at}"
+      raise OutOfOrderTransmissionError, error
```
elliotcm: This error message, while useful, gets discarded in the controller.

db/migrate/20151103170155_add_default_transmitted_at_to_content_items.rb
```diff
@@ -0,0 +1,5 @@
+class AddDefaultTransmittedAtToContentItems < Mongoid::Migration
+  def self.up
+    ContentItem.where(transmitted_at: nil).update_all(transmitted_at: 1)
```
elliotcm: Will this be reliably coerced to a float by mongoid?

spec/factories/content_item.rb
```diff
@@ -7,6 +7,7 @@
     format 'gone' # Using gone as it allows the smallest valid base
     publishing_app 'publisher'
     routes { [{ 'path' => base_path, 'type' => 'exact' }] }
+    transmitted_at 1
```
elliotcm: 1.0?

spec/integration/end_to_end_spec.rb
```diff
@@ -13,7 +13,8 @@
     "routes" => [
       { "path" => "/vat-rates", "type" => 'exact' }
     ],
-    "public_updated_at" => Time.now
+    "public_updated_at" => Time.now,
+    "transmitted_at" => 2,
```
elliotcm: 2.0?

spec/integration/submitting_redirect_item_spec.rb
```diff
@@ -11,7 +11,8 @@
       "redirects" => [
         {"path" => "/crb-checks", "type" => "prefix", "destination" => "/dbs-checks"},
         {"path" => "/crb-checks.json", "type" => "exact", "destination" => "/api/content/dbs-checks"}
-      ]
+      ],
+      "transmitted_at" => 2
```
elliotcm: Let's stick a comma on this to avoid the same noisy diff next time?

spec/models/content_item_spec.rb
```diff
@@ -34,6 +39,49 @@
...
+        it "upserts the item" do
+          result = item = nil
+          expect {
+            result, item = ContentItem.create_or_replace(@item.base_path, attributes)
+          }.not_to change(ContentItem, :count)
+          expect(result).to eq(:replaced)
+          expect(item.transmitted_at).to eq(2)
+        end
+      end
+
```
elliotcm: Whitespace.

elliotcm:
General approach looks good.  I'm concerned that we're using integers for the `transmitted_at` in almost all the tests but actually comparing floats in the real code.

bradwright:
Why are we using floats at all?

tuzz:
Addressed. We're now sending strings in the request payload and comparing with Ruby's BigNum integers, which are more reliable.

elliotcm:
@tuzz You'll need to update your pact tests to account for the timestamp change.

tuzz:
> @elliotcm @tuzz You'll need to update your pact tests to account for the timestamp change.

Already done – these tests are failing because it builds this branch against the master pact in publishing api (which doesn't exist yet). There's a dependency that the other app's pull request needs to be merged before this one.

PR #167, opened 2015-11-10
Add email_alert_type to email signup items

mobaig:
This will support the email-alert-frontend by providing it with the
information it needs to correctly construct a subscription request for
the email-alert-api.

### DIFF
--- a/db/migrate/20151110173144_add_signup_type_to_email_alert_signups.rb
+++ b/db/migrate/20151110173144_add_signup_type_to_email_alert_signups.rb
@@ -1,0 +1,15 @@
+class AddSignupTypeToEmailAlertSignups < Mongoid::Migration
+  def self.up
+    ContentItem.any_of({base_path: /.*government.*email-signup.*/, rendering_app: "email-alert-frontend"}).each do |policy_email_signup|
+      policy_email_signup.details[:email_alert_type] = "policies"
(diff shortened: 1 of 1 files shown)


rboulton:
I think this will probably do the right thing - though I'd want to see a list of what matches the `/.*government.*email-signup.*/` regular expression to be sure there are no other pages from other bits of the email system which match that pattern.  Perhaps checking that the rendering app is email-alert-frontend would be a good safety check.

More generally (though a non merge-blocking thought) - I wonder if we should avoid doing migrations like this in derived systems wherever possible.  An alternative approach would just be to re-register all the policies from policy-publisher, which would ensure that there's no risk of a difference between what policy publisher is sending now and what our migration produces.

I think we'd certainly want to avoid making migrations like this in the publishing-api, because then the publishing api's event log won't match its derived representations, such that replaying the event log to produce new derived representations would do the wrong thing.

mobaig:
@rboulton List of all matches from that query, with the rendering app printed on the next line: https://gist.github.com/mobaig/462f359c41feaa1ab354

Regarding the second point - that seems sensible. For now I can make sure that policy publisher gets deployed before this migration is run, but in principle re-registering all policies is a more sound approach.

mobaig:
@rboulton have also added a check for the rendering app in the ContentItem query.

PR #169, opened 2015-11-19
[Do Not Merge] Generate content ids when missing from ContentItem...

steventux:
Please don't merge yet as there are a couple of applications that still require updates to how content_ids are generated and tracked before we can update the remaining data.
Generate `content_id` from `SecureRandom.uuid` for content items which do
not have a value.
Copy the `updated_at` value to `public_updated_at` field for renderable
content items which don't have a value.

Reports updates to files in `./tmp`

### DIFF
--- a/db/migrate/20151119100304_update_blank_content_item_content_ids.rb
+++ b/db/migrate/20151119100304_update_blank_content_item_content_ids.rb
@@ -3,7 +3,7 @@ class UpdateBlankContentItemContentIds < Mongoid::Migration
     content_items = []
 
     ContentItem.where(content_id: nil).each do |item|
-      item.update!(content_id: SecureRandom.uuid)
+      item.set(content_id: SecureRandom.uuid)
 
       content_items << "#{item.content_id},#{item.publishing_app},#{item.base_path}"
     end
@@ -17,7 +17,7 @@ class UpdateBlankContentItemContentIds < Mongoid::Migration
     content_items = []
 
     ContentItem.renderable_content.where(public_updated_at: nil).each do |item|
-      item.update!(public_updated_at: item.updated_at)
+      item.set(public_updated_at: item.updated_at)
       content_items << "#{item.public_updated_at},#{item.publishing_app},#{item.base_path}"
     end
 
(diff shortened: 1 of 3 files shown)


db/migrate/20151119100304_update_blank_content_item_content_ids.rb
```diff
@@ -0,0 +1,33 @@
+class UpdateBlankContentItemContentIds < Mongoid::Migration
+  def self.up
+    content_items = []
+
+    ContentItem.where(content_id: nil).each do |item|
+      item.update!(content_id: SecureRandom.uuid)
```
elliotcm: If you like you can use `update_column` so it doesn't adjust the timestamps.

db/migrate/20151119100304_update_blank_content_item_content_ids.rb
```diff
@@ -0,0 +1,33 @@
+class UpdateBlankContentItemContentIds < Mongoid::Migration
+  def self.up
+    content_items = []
+
+    ContentItem.where(content_id: nil).each do |item|
```
elliotcm: Might wanna `find_each` as there's quite a few of these.

steventux: Mongoid doesn't implement `find_each` on `Mongoid::Criteria` apparently. Having read a bit about this I'm reasonably assured that the internals of `where().each` are using a cursor to iterate the collection. Further info here: http://stackoverflow.com/questions/7041224/finding-mongodb-records-in-batches-using-mongoid-ruby-adapter

db/migrate/20151119100304_update_blank_content_item_content_ids.rb
steventux: Good point, I've switched this to an atomic `set`

db/migrate/20151119100304_update_blank_content_item_content_ids.rb
elliotcm: Oh yes.  Mongo.

steventux:
https://github.com/alphagov/content-store/pull/175 supersedes this

PR #172, opened 2015-11-30
Content type presentation

tuzz:
https://trello.com/c/nkofgTdj/372-govspeak-part-2-move-contenttyperesolver-to-a-presenter-in-the-contentstore

This is the other half of the govspeak story. We decided that it would be better to move the content type resolution into the content store presentation layer.

This pull request includes a data migration that is quite slow (~8 minutes), so I've made sure that the implementation is backwards compatible whilst it is migrating.

We can remove that backwards compatibility in a separate pull request later.

The publishing api pull request is here: https://github.com/alphagov/publishing-api/pull/156

This pull request needs to be merged first.

### DIFF
--- a/app/models/content_item.rb
+++ b/app/models/content_item.rb
@@ -46,7 +46,7 @@ class ContentItem
   field :_id, :as => :base_path, :type => String, :overwrite => true
   field :content_id, :type => String
   field :title, :type => String
-  field :description, :type => String
+  field :description, :type => Hash, :default => { "value" => nil }
   field :format, :type => String
   field :locale, :type => String, :default => I18n.default_locale.to_s
   field :need_ids, :type => Array, :default => []
@@ -80,6 +80,16 @@ class ContentItem
     super(options).tap do |hash|
       hash["base_path"] = hash.delete("_id")
     end
+  end
+
+  # We store the description in a hash because Publishing API can send through
+  # multiple content types.
+  def description=(value)
+    super("value" => value)
+  end
+
+  def description
+    super.fetch("value")
   end
 
   def redirect?
--- a/app/models/content_item.rb
+++ b/app/models/content_item.rb
@@ -89,7 +89,16 @@ class ContentItem
   end
 
   def description
-    super.fetch("value")
+    description = super
+
+    if description.is_a?(Hash)
+      description.fetch("value")
+    else
+      # This is here to ensure backwards compatibility during data migration:
+      # db/migrate/20151130111755_description_value_hash.rb
+      # It can be removed afterwards.
+      description
+    end
   end
 
   def redirect?
--- a/app/presenters/content_type_resolver.rb
+++ b/app/presenters/content_type_resolver.rb
@@ -1,0 +1,44 @@
+class ContentTypeResolver
+  def initialize(content_type)
+    self.content_type = content_type
+  end
+
+  def resolve(object)
+    if object.is_a?(Hash)
+      resolve_hash(object)
+    elsif object.is_a?(Array)
+      resolve_array(object)
+    else
+      object
+    end
+  end
+
+private
+  def resolve_hash(hash)
+    hash.inject({}) do |hash, (key, value)|
+      hash.merge(key => resolve(value))
+    end
+  end
+
+  def resolve_array(array)
+    if has_content_types?(array)
+      extract_content(array)
+    else
+      array.map { |e| resolve(e) }
+    end
+  end
+
+  def has_content_types?(array)
+    return false unless array.all? { |v| v.is_a?(Hash) }
+
+    hash_keys = array.flat_map(&:keys).map(&:to_sym)
+    hash_keys.include?(:content_type)
+  end
+
+  def extract_content(array)
+    array = array.map(&:symbolize_keys)
+    array.detect { |h| h[:content_type] == content_type }[:content]
+  end
+
+  attr_accessor :content_type
+end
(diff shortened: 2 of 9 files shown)


app/models/content_item.rb
```diff
@@ -82,6 +82,25 @@ def as_json(options = nil)
...
+
+  def description
+    description = super
+
+    if description.is_a?(Hash)
+      description.fetch("value")
+    else
+      # This is here to ensure backwards compatibility during data migration:
+      # db/migrate/20151130111755_description_value_hash.rb
+      # It can be removed afterwards.
```
elliotcm: Have to taken into consideration the fact that the database change happens before the services are restarted?  I don't know how this affects migrations with mongo.

app/models/content_item.rb
```diff
@@ -46,7 +46,7 @@ def self.create_or_replace(base_path, attributes)
   field :_id, :as => :base_path, :type => String, :overwrite => true
   field :content_id, :type => String
   field :title, :type => String
-  field :description, :type => String
+  field :description, :type => Hash, :default => { "value" => nil }
```
elliotcm: This will change the effective default from empty string to nil, right?

tuzz: This sets the default to be a hash containing `{ "value" => nil }`, which is consistent with the approach we took in publishing api.

When you call `#description` on the model it "unpacks" the value and returns the nested string (or array in the case of there being content types).

app/models/content_item.rb
tuzz: Can you elaborate more? Does that means the application's running the old code whilst migrations are running? That's not going to work, if that's the case.

app/models/content_item.rb
elliotcm: Yes, but where previously calling `description` on a new object would've resulted in a blank string, does it not now result in a `nil`?

app/models/content_item.rb
elliotcm: The migration itself is running in the context of the new code (because it's a new `rake` process) but the web workers aren't restarted until after the migration has finished.  That's more of a problem with relational DBs because fixed schemas, but I dunno how mongo(id) would react.

app/presenters/content_type_resolver.rb
```diff
@@ -0,0 +1,45 @@
...
+    if object.is_a?(Hash)
+      resolve_hash(object)
+    elsif object.is_a?(Array)
+      resolve_array(object)
+    else
+      object
+    end
+  end
+
+  private
```
elliotcm: Outdent plz

app/presenters/content_type_resolver.rb
```diff
@@ -0,0 +1,45 @@
...
+  def has_content_types?(array)
+    return false unless array.all? { |v| v.is_a?(Hash) }
+
+    hash_keys = array.flat_map(&:keys).map(&:to_sym)
+    hash_keys.include?(:content_type)
+  end
+
+  def extract_content(array)
+    array = array.map(&:symbolize_keys)
+    array.detect { |h| h[:content_type] == content_type }[:content]
```
elliotcm: Have we validated by this point that the `content` key exists?  I know it *should* in theory.

db/migrate/20151130111755_description_value_hash.rb
```diff
@@ -0,0 +1,19 @@
...
+    puts "'some description' : { 'value' => 'some description' }"
+
+    ContentItem.all.each.with_index do |item, index|
+      print "." if (index % 100).zero?
+
+      description = item["description"]
+
+      unless description.is_a?(Hash)
+        item.description = description
+        item.save!(validate: false)
```
elliotcm: If we don't want validations perhaps `update_column` would be better, since that won't adjust the `updated_at` either (which is a fallback for public updated at, I think).

elliotcm:
Broadly :+1:

app/models/content_item.rb
tuzz: It will now result in `nil`, yes.

Did it return a blank string before? I assumed that because there was no default, it would return `nil`, but I'll check.

app/models/content_item.rb
tuzz: I think it will be inconsistent until the web workers are restarted. Perhaps this needs to be broken into two deploys:

1) Deploy the backwards compatible code
 (web workers restarted)
2) Migrate the data

What do you think?

app/presenters/content_type_resolver.rb
tuzz: Yeah, there's a validator in Publishing API here: https://github.com/alphagov/publishing-api/blob/forward-html-content-type-only/app/models/draft_content_item.rb#L38-L39

We could check again in content store, but we're trying to do as much validation "upfront", as per the previous work to move the validations into the publishing api.

db/migrate/20151130111755_description_value_hash.rb
tuzz: That was my first try, but it's not a thing in Mongoid:

```
ContentItem.first.update_column
NoMethodError: undefined method `update_column' for #<ContentItem:0x007fc541924998>
```

app/models/content_item.rb
tuzz: It looks like having a `String` type field sets defaults the value to `nil`:

```ruby
ContentItem.create.title
#=> nil
```

So this should be ok?

tuzz:
Comments addressed @elliotcm

app/models/content_item.rb
tuzz: I have reordered the commits such that the data migration happens last. This means that we could deploy the migration separately.

elliotcm:
Agree, let's deploy these in two steps.

tuzz:
:+1:

PR #174, opened 2015-12-02
Add endpoint for an item's incoming links

tijmenb:
The `/incoming-links` endpoint can be used to get the "incoming" links for a content item.

It will be used by the collections to display a list of pages tagged to a particular topic.

There was some discussion beforehand about this, see RFC 37:

https://gov-uk.atlassian.net/wiki/display/FS/RFC+37%3A+Add+linked+items+to+content+items#

Related Trello ticket: https://trello.com/c/bsZAwZ35

### DIFF
--- a/app/presenters/linked_item_presenter.rb
+++ b/app/presenters/linked_item_presenter.rb
@@ -26,7 +26,6 @@ class LinkedItemPresenter
   end
 
 private
-
   def api_url(item)
     @api_url_method.call(item.base_path_without_root)
   end
--- a/app/presenters/linked_item_presenter.rb
+++ b/app/presenters/linked_item_presenter.rb
@@ -1,0 +1,37 @@
+# Presenter for generating the public-facing representation of content item links.
+class LinkedItemPresenter
+  attr_reader :linked_item, :api_url_method
+
+  def initialize(linked_item, api_url_method)
+    @linked_item = linked_item
+    @api_url_method = api_url_method
+  end
+
+  def present
+    presented = {
+      "content_id" => linked_item.content_id,
+      "title" => linked_item.title,
+      "base_path" => linked_item.base_path,
+      "description" => ContentItemPresenter::RESOLVER.resolve(linked_item.description),
+      "api_url" => api_url(linked_item),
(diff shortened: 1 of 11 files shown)


app/presenters/linked_item_presenter.rb
```diff
@@ -0,0 +1,37 @@
...
+
+    if linked_item.has_attribute?(:analytics_identifier)
+      presented["analytics_identifier"] = linked_item.analytics_identifier
+    end
+
+    presented
+  end
+
+private
+
```
benlovell: This line should go according to the style guide.

spec/integration/fetching_linked_items_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+
+      expect(parsed_response["parent"]).to eql([
+        {
+          "content_id" => "ID-1",
+          "title" => "A",
+          "base_path" => "/a",
+          "description" => nil,
+          "api_url" => "http://www.example.com/content/a",
+          "web_url" => "https://www.test.gov.uk/a",
+          "locale"=>"en"
```
benlovell: Missing spaces around the hashrocket.

spec/integration/fetching_linked_items_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+          "title" => "B",
+          "base_path" => "/b",
+          "description" => nil,
+          "api_url" => "http://www.example.com/content/b",
+          "web_url" => "https://www.test.gov.uk/b",
+          "locale" => "en"
+        },
+      ])
+
+      expect(parsed_response["topics"]).to eql([])
```
benlovell: Maybe `be_empty` is clearer than `eql([])`?

tijmenb: No, that would mean that if `parsed_response["topics"]` returned a hash, this test would also pass. `{}.empty?` is also true.

benlovell: That's true (and very precise)

benlovell:
:+1:

PR #176, opened 2015-12-07
Data cleanup

tuzz:
These are some rake tasks that we intend to use for the data cleanup work where content store will be set to a valid state, i.e. all content items have content_ids and public_updated_at timestamps. We will then export this data to a JSON file and import it into the Publishing API.

### DIFF
--- a/lib/tasks/data_hygiene/missing_attributes.rake
+++ b/lib/tasks/data_hygiene/missing_attributes.rake
@@ -1,0 +1,61 @@
+namespace :data_hygiene do
+  def write_file(report_path, content_items)
+    File.open(report_path, "w") do |file|
+      file.puts "content_id,public_updated_at,publishing_app,base_path"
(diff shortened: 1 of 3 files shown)


lib/tasks/data_hygiene/missing_attributes.rake
```diff
@@ -0,0 +1,61 @@
...
+  end
+
+  desc "Assigns a public_updated_at for content items that do not have one"
+  task assign_public_updated_at: :environment do
+    content_items = []
+
+    ContentItem.where(public_updated_at: nil).each do |item|
+      item.set(public_updated_at: item.updated_at)
+
+      content_items << "#{item.public_updated_at},#{item.public_updated_at},#{item.publishing_app},#{item.base_path}"
```
danielroseman: Think the first column should be item.content_id.

PR #178, opened 2015-12-08
Fix content_id mismatches between draft and live content stores

tuzz:
**TL;DR** Make mismatching content_ids match in both content stores.

We have content items for the same `base_path` in both content stores which do not have matching `content_id`s, this could be the result of formats changing, in most cases this appears to be a format change like `placeholder_specialist_publication` changing to `specialist_publication`, the theory being that if the draft has been updated to a `gone` or `redirect` then modified to the new format the `content_id` will have changed on the draft content store for the given `base_path`.

There are also a handful of cases where non-english locale suffixes are not present on `base_path`, this PR also adds a task to handle this.

### DIFF
--- a/lib/tasks/data_hygiene/draft_content_id_cleanup.rake
+++ b/lib/tasks/data_hygiene/draft_content_id_cleanup.rake
@@ -1,0 +1,47 @@
+namespace :data_hygiene do
+  namespace :draft_content_id_cleanup do
+    def report(cleanup: false)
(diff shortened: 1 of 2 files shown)


jamiecobbett:
For people like myself who don't know why there are mismatches, could you summarise how they got there?

tuzz:
@jamiecobbett We updated the PR description above. We can clarify further if needed. Let us know.

PR #182, opened 2016-01-25
Pass through links hashes

elliotcm:
This allows the publishing API to provide
"hard-coded" content for linked items in the cases
where we cannot currently represent them as dynamic
items in the publishing API, for example governments
(for history mode).

This is a temporary measure until a publishing API
refactor takes over the expansion of these links
hashes.

https://trello.com/c/ZahFZJR3/510-pass-through-expanded-links

### DIFF
--- a/app/models/content_item.rb
+++ b/app/models/content_item.rb
@@ -131,6 +131,8 @@ class ContentItem
   end
 
   def base_path_without_root
+    return nil unless base_path
+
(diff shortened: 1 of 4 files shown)


danielroseman:
This could do with some integration specs, to verify that the endpoint responds correctly when given a mixture of link types.

elliotcm:
@danielroseman updated.

PR #183, opened 2016-01-26
Upgrade Rails and rails-html-sanitizer.

danielroseman:
http://weblog.rubyonrails.org/2016/1/25/Rails-5-0-0-beta1-1-4-2-5-1-4-1-14-1-3-2-22-1-and-rails-html-sanitizer-1-0-3-have-been-released/


bradwright:
Merged to ensure security, but do we have a plan for going to `4.2.x`?

PR #188, opened 2016-02-10
Expand extra fields for Topical Events in links hash

boffbowsh:
The start and end dates for Topical Events are needed to render Topical
Event About Pages. We expand these fields in the links hash to simplify
the logic in the frontend.

We need to remove the returned field filter for linked items in order
to achieve this, as changing these `only` lines for each format is not
sustainable.

We benchmarked the removal of the `only` lines from the queries, and also
tried using `without("details.body")` which is apparently slower (thanks, Mongo(id)?).

```
Benchmark.bm(15) do |x|
  x.report("only") { 1000.times { ContentItem.only(:content_id, :locale, :base_path, :title, :description).first.to_json } }
  x.report("without body") { 1000.times { ContentItem.without("details.body").first.to_json } }
  x.report("all") { 1000.times { ContentItem.first.to_json } }
end

                      user     system      total        real
only              0.930000   0.030000   0.960000 (  1.144666)
without body      1.330000   0.030000   1.360000 (  1.567574)
all               1.130000   0.040000   1.170000 (  1.369165)
```

/cc @danielroseman

### DIFF
--- a/app/models/content_item.rb
+++ b/app/models/content_item.rb
@@ -158,7 +158,6 @@ private
       .renderable_content
       .where(:content_id => {"$in" => content_ids})
       .where(:locale => {"$in" => [I18n.default_locale.to_s, self.locale].uniq})
-      .only(:content_id, :locale, :base_path, :title, :description, :analytics_identifier)
       .sort(:updated_at => -1)
       .group_by(&:content_id)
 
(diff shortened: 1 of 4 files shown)


benlovell:
General aside: I tend to use https://github.com/evanphx/benchmark-ips, as you can configure warmup and it shows a standard deviation, just so you can be sure you don't have some odd variability creeping in.

gpeng:
Happy to merge unless you're waiting on a 👍from @danielroseman

boffbowsh:
I am indeed, I should have mentioned that. Or at least someone on his team.

jamiecobbett:
:+1:

PR #189, opened 2016-02-12
Remove UpdateLock and transmitted_at

gpeng:
Due to time differences across the host machines of +/- 3ms this
approach is error prone once deployed to integration/staging/production.
This commit just removes the `transmitted_at` field and `UpdateLock`. It
doesn't currently `unset` `transmitted_at` in Mongo. I've not added an
alternative locking approach here as that may well be better added in
the Publishing API.

We've had some debate about whether this is feasible at this stage so I've opened the PR for comments and discussion. 

The pact tests are currently failing as I've not updated them in Publishing API


tuzz:
Simpler is better. :+1:

jamiecobbett:
Reading the [original](https://github.com/alphagov/content-store/pull/166) [PRs](https://github.com/alphagov/publishing-api/pull/127) for adding this in, the intent was to make sure that content-store ended up with the most recent version of an item. @gpeng's [changes in publishing-api](https://github.com/alphagov/publishing-api/pull/203) now mean that publishing-api should always be sending the latest version anyway now.

So, :+1:

boffbowsh:
Closing in favour of a new atomic counter

PR #193, opened 2016-02-24
Support `payload_version` in `ContentItem`

gpeng:
This PR adds support for `payload_version` without yet deprecating `transmitted_at`. This will allow us to update Publishing API to start sending `payload_version` after which `transmitted_at` can be removed.

`payload_version` is essentially a locking mechanism between Publishing API and Content Store to prevent out of order 'old' content overwriting a newer version. This can be caused during the asynchronous queueing of payloads on their way from Publishing API. It will be set on the content item payloads as they are queued in Publishing API.

I've not updated the pact tests as yet but will do that when `transmitted_at` is added to Publishing API and then removed from Content Store in subsequent PRs.

[Trello](https://trello.com/c/dBmI96gI/304-change-locking-criteria-on-content-store)

### DIFF
--- a/app/models/update_lock.rb
+++ b/app/models/update_lock.rb
@@ -2,28 +2,66 @@ class UpdateLock
   def initialize(lockable)
     @lockable = lockable
 
-    if lockable && !lockable.respond_to?(:transmitted_at)
-      raise ArgumentError.new("#{lockable.class} must implement transmitted_at")
+    unless object_is_lockable?(lockable)
+      raise ArgumentError.new("#{lockable.class} must implement transmitted_at or payload_version")
     end
   end
 
   def check_availability!(attributes)
-    transmitted_at = attributes.with_indifferent_access[:transmitted_at]
-    raise MissingAttributeError, "transmitted_at is mandatory" unless transmitted_at
     return unless lockable.present?
 
-    request_transmitted_at = Integer(transmitted_at)
-    database_transmitted_at = Integer(lockable.transmitted_at)
+    attributes = attributes.with_indifferent_access
 
-    if database_transmitted_at >= request_transmitted_at
-      error = "Tried to process request with transmitted_at #{request_transmitted_at},"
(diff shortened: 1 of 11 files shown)


app/models/update_lock.rb
```diff
@@ -3,27 +3,60 @@ def initialize(lockable)
...
+    attributes = attributes.with_indifferent_access
 
-    if database_transmitted_at >= request_transmitted_at
-      error = "Tried to process request with transmitted_at #{request_transmitted_at},"
-      error += " but the latest #{lockable.class} has a newer (or equal) transmitted_at of #{lockable.transmitted_at}"
-      raise OutOfOrderTransmissionError, error
-    end
+    validate_attributes(attributes)
+
+    payload_version, transmitted_at = attributes[:payload_version], attributes[:transmitted_at]
```
benlovell: You could still use parallel assignment with `attributes.values_at(:payload_version, :transmitted_at)`

spec/models/update_lock_spec.rb
```diff
@@ -17,41 +17,111 @@
 
       it "does not raise an error" do
         expect {
-          subject.check_availability!(2)
+          subject.check_availability!({transmitted_at: "1"})
```
benlovell: I don't think the braces around the `transmitted_at` option are strictly necessary.

boffbowsh:
Mostly :+1:, just some small comments

app/models/update_lock.rb
gpeng: Fixed in e523571

spec/models/update_lock_spec.rb
gpeng: Fixed in e523571

gpeng:
@boffbowsh Should be fixed now (mostly in e523571 I think but your comments got 🍳)

boffbowsh:
:+1: deploy to integration to test?

PR #194, opened 2016-02-26
Include specific incoming links in content

jamiecobbett:
The interesting change is in the last commit.

To render a working_group, you need the list of policies which link to that
working_group.

We could fulfill that by adding special knowledge to the frontend so that it
knows to make a second request, to /incoming-links. There are a few issues with
that:
* one of the design goals of the content-store is that it includes everything
needed to render that page
* it complicates the frontend
* we think that when dependency resolution is implemented, the API of
content-store will change to match this implementation, and that the
/incoming-links endpoint could be removed

If the API design does change in this way, then we don't need to do anything to
our frontends when we implement dependency resolution.

I'd rather it wasn't so directly hardcoded - instead having a list with an
entry for each format, containing the various strings - but I had difficulty
with naming. Probably best to refactor when we get to the second or third use
case.

We are reasonably confident that there will be other formats where this
functionality will be required.

I see this change as being similar to 04368e5.

### DIFF
--- a/app/queries/linked_items_query.rb
+++ b/app/queries/linked_items_query.rb
@@ -1,0 +1,78 @@
+class LinkedItemsQuery
+  attr_reader :content_item
+
+  def initialize(content_item)
+    @content_item = content_item
+  end
+
+  def call
+    items = linked_items
+    items["available_translations"] = available_translations if available_translations.any?
+    items
+  end
+
+private
+
+  # Return the extant and renderable linked_items and any "passthrough" items
+  # which aren't really content_items, but are instead hardcoded hashes.
+  def linked_items
+    outgoing_content_items = content_items
+    content_item.links.each_with_object({}) do |(link_type, raw_content_ids), result|
+      passthrough_hashes, raw_content_ids = raw_content_ids.partition { |link| link.is_a?(Hash) }
+      passthrough_content_items = passthrough_hashes.map do |attributes|
+        ContentItem.new(attributes)
+      end
+
+      real_content_items = raw_content_ids.map do |id|
+        outgoing_content_items.find { |ci| ci.content_id == id }
+      end
+      real_content_items.compact!
+      result[link_type] = real_content_items + passthrough_content_items
+    end
+  end
+
+  # Return the renderable items based on `content_ids`, in the most appropriate
+  # locale
+  def content_items
+    # For each linked content_id find all non-redirect content items with
+    # matching content_id in either this item's locale, or the default locale
+    # with the most recently updated first.
+    locales = [I18n.default_locale.to_s, content_item.locale].uniq
+    renderable_items_by_id = ContentItem
+      .renderable_content
+      .where(content_id: { "$in" => all_content_ids })
+      .where(locale: { "$in" => locales })
+      .sort(updated_at: -1)
+      .group_by(&:content_id)
+
+    # For each set of items for a given content_id, pick the first one that
+    # matches this item's locale, or fall back to the first one matching the
+    # default locale.
+    renderable_items_by_id.map do |_content_id, items|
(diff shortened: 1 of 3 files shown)


app/queries/linked_items_query.rb
```diff
@@ -0,0 +1,84 @@
...
+
+      real_content_items = raw_content_ids.map { |id| content_items_by_id[id] }.compact
+      result[link_type] = real_content_items + passthrough_content_items
+    end
+  end
+
+  # Return the renderable items based on `content_ids`, in the most appropriate
+  # locale
+  def content_items_by_id
+    defined?(@_content_items) ? @_content_items : @_content_items = begin
```
boffbowsh: :+1: like this memoization pattern. Nifty!

jamiecobbett: It avoids accidental recomputation if the return value is falsey, which is a problem with:
```ruby
@_content_items ||= ...
```

boffbowsh:
LGTM :+1:

tijmenb:
I'm not sure about the approach. It seems similar to what [RFC 37](https://gov-uk.atlassian.net/wiki/display/FS/RFC+37%3A+Add+linked+items+to+content+items) is about. @bradwright says

> "As part of this we recommend that we don't constrain our thinking to single API calls for all pages - there are a small amount of very complex pages on GOV.UK that we'll always be unable to easily render in a single call, and we should endeavour to provide the right tools to support more complex pages."

The main thing probably is that if a working group page frontend is going to change, we'd have to change this endpoint too. How do we feel about that?

jamiecobbett:
@tijmenb I agree, but our rationale for doing it this way is that it's a fudge while we wait for dependency resolution. I think I'm right in saying that dependency resolution was something that came along after that RFC. 

This change feels acceptable to me for the same reason as https://github.com/alphagov/content-store/commit/04368e52d22d1e5422608ce18cf11e306014f377

> The main thing probably is that if a working group page frontend is going to change, we'd have to change this endpoint too. How do we feel about that?

In terms of dependency resolution, I think the plan was to express this configuration in the content-schemas as a way of constraining complexity and keeping it implementation-agnostic.

tijmenb:
Allright, understood! Looking forward to the dependency resolution, curious to see how that will turn out.

app/queries/linked_items_query.rb
```diff
@@ -0,0 +1,84 @@
...
+
+  def initialize(content_item)
+    @content_item = content_item
+  end
+
+  def call
+    items = linked_items
+    items["available_translations"] = available_translations if available_translations.any?
+    if content_item.format == "working_group"
+      items["policies"] = content_items_by_id.values.select { |content_item| content_item.format == "policy" }
```
jamiecobbett: There are two issues with this line as it stands:
* it would overwrite any policies that the item happened to link to
* it would include *any* policy returned by `content_items_by_id`, not the ones linking to the `working_group`

danielroseman:
:+1:, subject to the comments Jamie makes on LinkedItemsQuery.

app/queries/linked_items_query.rb
tijmenb: As for the overwriting issue, I would put the links in a different key in the content item:

Something like:

```
{
  content_id: "item-b",
  links: { ... },
  incoming_links: { policies: ["item-a"] }
}
```

jamiecobbett: I've resolved both the issues I've described above in a rebase.

@tijmenb and I discussed his suggestion offline and our conclusion was that the frontend shouldn't know that these links are different, so using `"links"` is the right thing to do.

jamiecobbett:
OK, I think this is now ready for review/merge.

boffbowsh:
:clap:

dsingleton:
:clap: :sparkles: :100:

PR #198, opened 2016-03-29
Respect content_item.max_cache_time if one is set...

tuzz:
https://trello.com/c/upCBCURU/635-spike-investigate-clearing-the-frontend-cache-more-quickly-1-day

**Respect content_item.max_cache_time if one is set...**

This also includes some tidying up of the controller
to consolidate the caching behaviour into one place.
Previously, it was spread out between methods and
the #show method which made it harder to understand.

**Track publish intent retrieval in statsd...**

This includes a rename to the statsd key for
looking up the content item for consistency.

### DIFF
--- a/app/controllers/content_items_controller.rb
+++ b/app/controllers/content_items_controller.rb
@@ -2,11 +2,13 @@ class ContentItemsController < ApplicationController
   before_filter :parse_json_request, only: [:update]
 
   def show
-    item = Rails.application.statsd.time('show.find_by') do
+    item = Rails.application.statsd.time('show.find_content_item') do
       ContentItem.where(base_path: encoded_base_path).first
     end
 
-    intent = PublishIntent.where(base_path: encoded_base_path).first
+    intent = Rails.application.statsd.time('show.find_publish_intent') do
+      PublishIntent.where(base_path: encoded_base_path).first
+    end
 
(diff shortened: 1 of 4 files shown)


app/controllers/content_items_controller.rb
```diff
@@ -1,14 +1,22 @@
 class ContentItemsController < ApplicationController
   before_filter :parse_json_request, only: [:update]
-  before_filter :set_default_cache_headers, only: [:show]
 
   def show
-    item = Rails.application.statsd.time('show.find_by') do
-      ContentItem.find_by(base_path: encoded_base_path)
+    item = Rails.application.statsd.time('show.find_content_item') do
+      ContentItem.where(base_path: encoded_base_path).first
```
dougdroper: what's the reason behind changing from find_by to where?

tuzz: In Mongoid, a `#find_by` raises an error if the record doesn't exist (annoyingly inconsistent!). I've explicitly raised this error after the cache headers are set further down in this method. The reason to do this was so that I could consolidate the caching behaviour into a single method. Previously it was a bit tangled and scattered throughout the controller.

dougdroper: :+1:

PR #199, opened 2016-03-29
Move max_cache_time into details hash...

tuzz:
I mistakenly assumed it would be a top-level field
when I coded this up. I ended up placing this field
in the details hash.

### DIFF
--- a/app/controllers/content_items_controller.rb
+++ b/app/controllers/content_items_controller.rb
@@ -77,8 +77,8 @@ private
     elsif item && item.access_limited?
       cache_time = config.minimum_ttl
       is_public = false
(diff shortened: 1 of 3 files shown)


app/controllers/content_items_controller.rb
```diff
@@ -94,4 +94,11 @@ def bounded_max_age(cache_time)
       cache_time
     end
   end
+
+  def max_cache_time(item)
+    return unless item
+    return unless item.details
```
danielroseman: This could be simplified by using `try`.

tuzz:
@danielroseman changed to `#try`

PR #203, opened 2016-04-14
Update presenters to return nested parents.

tvararu:
This is required as part of migrating DetailedGuides. Once we have dependency resolution, we will be able to implement this more elegantly.

After this, all parents will export with a shape similar to:

```
{
  "links": {
    "parent": [
      {
        "content_id": "111-111",
        "title": "VAT Rates",
        "base_path": "/tax/vat-rates",
        "parent": [
          {
            "content_id": "222-222",
            "title": "Taxation",
            "base_path": "/tax",
            "parent": [
              {
                "content_id": "333-333",
                "title": "Home",
                "base_path": "/"
              }
            ]
          }
        ]
      }
    ]
  }
}
```

/cc @binaryberry.


app/presenters/linked_item_presenter.rb
```diff
@@ -22,6 +22,11 @@ def present
       presented[attr.to_s] = linked_item.send(attr) if linked_item.has_attribute?(attr)
     end
 
+    linked_topic = linked_item.linked_topic
```
boffbowsh: You can do this:
```ruby
if (linked_topic = linked_item.linked_topic)
  presented["links"][:topics][0] = add_topic_parent_recursively(linked_topic)
end
```

app/models/content_item.rb
```diff
@@ -138,6 +138,26 @@ def base_path_without_root
     base_path.sub(%r{^/}, "")
   end
 
+  # This method is temporary. It will be removed when dependency resolution is implemented.
+  # This method, when called on a ContentItem, will return the first Topic ContentItem that is linked
+  # or nil if none exists.
+  def linked_topic
+    topics = self[:links][:topics]
+    return if topics.nil?
+    linked_topic_content_id = topics.first[:content_id]
```
boffbowsh: Is `first` the correct behaviour? What about content tagged with multiple topics?

app/presenters/linked_item_presenter.rb
tvararu: Not a fan of assignments inside conditionals. I think it makes the code more clever, which I try to avoid.

boffbowsh: Fair enough 👍

tijmenb:
As discussed with offline at a few occasions, I thought it would've been preferrable to temporarily solve this with an extra call in client. We've done this in Whitehall in upcoming PR https://github.com/alphagov/whitehall/pull/2557. Our code: [specialist_tag_finder.rb#L23-L29](https://github.com/alphagov/whitehall/blob/23df09ed8ccc71adeec6401e8d1008db5c32f435/lib/specialist_tag_finder.rb#L23-L29)

However I may not have been part of all your discussions, so I'll roll with this.
 
But I'm not a 100% sure that the implementation is correct. This seems to take the first `topic`, and then recursively look up the parent for that topic. If you're going to use this is a breadcrumb, the idea is that we  recursively expand `links.parent` (which is of the `topic` type incidentally, but that's not important here). 

Or am I not understanding the use case? Perhaps the PR description could provide some why's & example output.

boffbowsh:
That's fine for Whitehall Frontend, but for new style Frontend apps we want them to only need to make a single call to content store to get everything they need to render the page.

This emulates the eventual behaviour of dependency resolution, it's essentially a content store-side shim in preparation for the dependency resolution work. In theory, when that's ready we'll be able to remove this shim and the frontend will continue to work in exactly the same way.

I'll let @tvararu explain the implementation more, but my rough understanding is that parent is used for breadcrumbs and the topics are used for the "Part of:" metadata. There is an open question around the breadcrumbs though - the parent of this document is set in Whitehall admin as "PAYE", however the parent of PAYE, "Business tax", is actually displayed as the breadcrumb. Do you happen to know why this is, @tijmenb? If not, we can try and investigate.

tvararu:
This implementation is dodgy, yes. I'm going to go back to Whitehall to re-trace how exactly it's rendering the topic parent when the model itself only has associations for the current topics. I've done this before but the process was non-trivial and I decided to back off.

tijmenb:
I'm not opposed to shims. The thinking here was that we may not be completely sure about how to represent these things in the content-store. Changing an API is not trivial, so we run the risk of committing to a less-than-ideal structure here. We chose to add the hack in Whitehall so that whenever dependency resolution comes around we'd have to change at most one thing. 

But let's leave that approach for now and focus on making the shim nice. Also because I didn't account for the "Part of:" content.

I'm off to a meeting now, but I'll try to figure out today 1) which data we currently show on these pages, 2) if it can be simplified 3) how we should store this in the `links` and 4) how we would want it back from the content store.

boffbowsh:
Thanks :). For record: that structure was suggested by @tuzz, based on the current plan for Dependency Resolution.

app/models/content_item.rb
tvararu: Discussed offline, but I'm fairly certain it's not. If the implementation remains as such, we would have to change it to apply to all of the linked topics.

tijmenb:
Do we have a record of that somewhere? https://github.com/alphagov/govuk-content-schemas/pull/279#discussion_r58688765 seems to suggest a different data structure.

tvararu:
The base key is different (`expanded_links.parent`), which is likely incorrect, but the rest is similar, no? If not, we should bring it in-line.

The tests on this branch expect:

```ruby
topics: [
  content_id: topic_paye_details.content_id,
  title: topic_paye_details.title,
  base_path: topic_paye_details.base_path,
  parent: [
    {
      content_id: topic_paye.content_id,
      title: topic_paye.title,
      base_path: topic_paye.base_path,
      parent: [
        {
          content_id: topic_business_tax.content_id,
          title: topic_business_tax.title,
          base_path: topic_business_tax.base_path
        }
      ]
    }
  ]
]
```

tvararu:
> There is an open question around the breadcrumbs though - the parent of this document is set in Whitehall admin as "PAYE", however the parent of PAYE, "Business tax", is actually displayed as the breadcrumb. Do you happen to know why this is, @tijmenb? If not, we can try and investigate.

In this comment I will attempt to answer that question from a technical standpoint to the best of my understanding. The particular detailed guide I am tracing through the stack is <https://www.gov.uk/guidance/salary-sacrifice-and-the-effects-on-paye>.

This displays the entire block in the frontend, where the "Business tax" link is present:

```ruby
<!-- views/detailed_guides/show -->
<header class="block headings-block">
  <div class="inner-block floated-children">
    <%= render  partial: 'header',
                locals: {
                  header_title: t_display_type(@document),
                  document: @document,
                  footer_meta: true,
                  policies: @related_policies,
                } %>
  </div>
</header>
```

Going deeper:

```ruby
<!-- views/documents/_header -->
<%
  header_title ||= ""
  policies ||= nil
  specialist_tag_finder = SpecialistTagFinder.new(@document)
%>
<%= render partial: 'shared/heading',
          locals: { type: add_sector_name(header_title, specialist_tag_finder.primary_sector_tag),
                    heading: document.title,
                    extra: true } %>
```

I'll skip some detours here: the `type` local being passed into `shared/heading` is the final HTML that gets presented. `add_sector_name` simply calls `specialist_tag_finder.primary_sector_tag.title` to get i.e. the `Business tax` string and appends ` - guidance` (`header_title`) to it.

So, what we need to know is what is `specialist_tag_finder.primary_sector_tag`:

```ruby
# lib/specialist_tag_finder.rb
class SpecialistTagFinder
  def initialize(document)
    @document = document
  end

  def primary_sector_tag
    primary_subsector_tag.parent if primary_subsector_tag
  end

  def primary_subsector_tag
    if primary_tag_slug = @document.primary_specialist_sector_tag
      specialist_sector_tags.find {|t| t.slug == primary_tag_slug }
    end
  end

private

  def artefact
    @artefact ||= Whitehall.content_api.artefact(RegisterableEdition.new(@document).slug)
  end

  def specialist_sector_tags
    return [] if artefact.nil?
    artefact.tags.select {|t| t.details['type'] == 'specialist_sector' }
  end
end
```

So, you call `primary_sector_tag` which calls `primary_subsector_tag` which will filter `specialist_sector_tags` based on the `@document.primary_specialist_sector_tag`. Which is:

```ruby
# models/edition/specialist_sectors.rb
def primary_specialist_sector_tag
  primary_specialist_sectors.first.try(:tag)
end
```

Note the `#first`. Which in the case of our Salary Sacrifice example is:

```ruby
> dg.primary_specialist_sectors.first
=> #<SpecialistSector id: 47465, edition_id: 595175, tag: "business-tax/paye", created_at: "2016-02-18 09:02:19", updated_at: "2016-02-18 09:02:19", primary: true>
> dg.primary_specialist_sectors.first.tag
=> "business-tax/paye"
```

Going back to `SpecialistTagFinder`, we're filtering `specialist_sector_tags` based on if their slug matches the above slug. What are `specialist_sector_tags` again?

```ruby
# Repeat from lib/specialist_tag_finder.rb
def artefact
  @artefact ||= Whitehall.content_api.artefact(RegisterableEdition.new(@document).slug)
end

def specialist_sector_tags
  return [] if artefact.nil?
  artefact.tags.select {|t| t.details['type'] == 'specialist_sector' }
end
```

So, we're querying the content store for a ContentItem that matches the current document's slug, which is assigned as our `@artefact`.

Well, that's what I hope it does anyway, because I can't get it to work in the Rails console possibly due to localisation magic:

```ruby
> re = RegisterableEdition.new(d)
=> #<RegisterableEdition:0x007fd277d7b260 @edition=#<Document id: 235854, created_at: "2014-05-15 14:18:48", updated_at: "2016-02-18 10:37:00", slug: "salary-sacrifice-and-the-effects-on-paye", document_type: "DetailedGuide", content_id: "5f58d036-7631-11e4-a3cb-005056011aef">>
> re.slug
NoMethodError: undefined method `non_english_edition?' for #<Document:0x007fd2780ffe90>
```

Within the `artefact.tags`, we filter out to only leave the ones that have a type that matches `specialist_sector`.

These tags will be again filtered, as aforementioned, based on if their slug matches `business-tax/paye`.

To be continued, will edit comment once I get past the `non_english_edition?` block.

boffbowsh:
Useful sleuthing so far, thanks!

tvararu:
I can't get past the localisation block, but I think the picture is clear now. The *first* `primary_specialist_sector` (also known as a `parent` in the new world) is what informs the choice of breadcrumb link and text.

Ergo, I'm going to update this to no longer decorate the first `topic`, but the `parent`.

tvararu:
This PR has been updated to now return the nested parent hierarchy.

app/presenters/linked_item_presenter.rb
```diff
@@ -44,4 +49,17 @@ def web_url(item)
 
     Plek.current.website_root + item.base_path
   end
+
+  def add_parents_recursively(parent)
+    json = {
+      content_id: parent.content_id,
+      title: parent.title,
+      base_path: parent.base_path
```
tvararu: Not sure about these 3 fields. Note that the frontend potentially needs more: https://github.com/alphagov/govuk-content-schemas/pull/279/commits/f2b26495258a990482d90b47c6633fae635f6618.

What's a cleaner way to get the "parent" representation of this content item?

boffbowsh: It'd be better to create a new instance of this class recursively for each ancestor.

app/presenters/linked_item_presenter.rb
```diff
@@ -22,6 +22,11 @@ def present
       presented[attr.to_s] = linked_item.send(attr) if linked_item.has_attribute?(attr)
     end
 
+    linked_parent = linked_item.linked_parent
+    if linked_parent
+      presented["links"][:parent][0] = add_parents_recursively(linked_parent)
```
boffbowsh:
```ruby
presented["links"][:parent][0] = self.class.new(linked_parent).present
```
or similar

tvararu:
@boffbowsh I've updated it to now use the suggested `self.class.new` invocation, which is a lot cleaner. 👍 

However, the structure has now slightly changed.

### Before (extra keys omitted for brevity)

```ruby
parent: [
  content_id: topic_paye_details.content_id,
  parent: [
    {
      content_id: topic_paye.content_id,
      parent: [
        {
          content_id: topic_business_tax.content_id,
        }
      ]
    }
  ]
]
```

### After

```ruby
links: {
  parent: [
    content_id: topic_paye_details.content_id,
    links: {
      parent: [
        {
          content_id: topic_paye.content_id,
          links: {
            parent: [
              {
                content_id: topic_business_tax.content_id,
                links: {}
              }
            ]
          }
        }
      ]
    }
  ]
}
```

Which is probably better. We'll have to update the schema.

boffbowsh:
Still WIP?

tvararu:
Actually, nope.

boffbowsh:
I think it's worth deploying this on integration first and checking we don't have a performance regression.

tvararu:
Agreed. 👍

tvararu:
Deployed to integration. Seems to work, but too well / not well at all, at the same time.

Given an object of this shape:

```json
"content_id": "1"
"links": {
  "parent": [{
    "content_id": "2",
    "links": {
      "parent": [
        "3"
      ]
    }
  }]
},
"topics: [{
  "content_id": "4"
  "links": {
    "parent": [
      "5"
    ]
  }
}]
```

It will expand **all** parents, not just the one we care about, which is at the root level link. Ergo:

```json
"content_id": "1"
"links": {
  "parent": [{
    "content_id": "2",
    "links": {
      "parent": [{
        "content_id": "3", // This is good.
        "links": {}
      }]
    }
  }]
},
"topics: [{
  "content_id": "4",
  "links": {
    "parent": [{
      "content_id": "5", // This is not good.
      "links": {} 
    }]
  }
}]
```

Which in retrospect makes perfect sense given how the code is written and what it does, but not exactly what we want.

Additionally, it appears that the content store lookups are botching up in some way that needs investigating.

Before this change: http://sprunge.us/QBgY?json

After this change: http://sprunge.us/BLDH?json

It's quite peculiar. Before, we have this (shortened, IDs replaced with shorter labels):

```json
"parent": [
  {
    "content_id": "A",
    "title": "PAYE",
    "links": {
      "parent": [
        "B"
      ]
    }
  }
],
"topics": [
  {
    "content_id": "A",
    "title": "PAYE",
    "links": {
      "parent": [
        "B"
      ]
    }
  }
],
"available_translations": [
  {
    "content_id": "C",
    "title": "Salary sacrifice",
    "links": {
      "parent": [
        "A"
      ]
    }
  }
]
},
```

So, we have a parent and a topic of `A`, which both point to `B` as their parent. Meanwhile, available translation `C` points to `A` as its parent.

Here's what actually happens:

```json
"parent": [
  {
    "content_id": "A",
    "title": "PAYE",
    "links": {
      "parent": [
        {
          "content_id": null, // What?!
          ...
          "schema_name": "gone",
          "document_type": "gone",
          "links": {}
        }
      ]
    }
  }
],
"topics": [
  {
    "content_id": "A",
    "title": "PAYE",
    "links": {
      "parent": [
        {
          "content_id": null, // Same one, here it is again.
          ...
          "schema_name": "gone",
          "document_type": "gone",
          "links": {}
        }
      ]
    }
  }
],
"available_translations": [
  {
    "content_id": "C",
    "title": "Salary sacrifice",
    "links": {
      "parent": [
        {
          "content_id": "A", // This one is good.
          "title": "PAYE",
          "links": {
            "parent": [
              {
                "content_id": "B", // Why is this one correct while the above link to garbage?!
                "title": "Business tax",
                "links": {
                  "children": ["etc"]
                }
              }
            ]
          }
        }
      ]
    }
  }
]
```

So yes. Need to:

1. Stop it from nesting parents of everything, and only nest `parent` `parent`s.
2. Figure out why it's linking to garbage data.

tijmenb:
Perhaps that the actual feature in publishing-api will be finished soon, and this temporary solution won't be necessary anymore. Have you checked with @dougdroper / @danielroseman?

tvararu:
I haven't, this has been sitting around for a while, but is nonetheless necessary for the detailed guidance migration (which has had a myriad of other loose ends to tie).

Doug / Daniel?

tvararu:
Had a chat with @dougdroper that explained the new `expanded_links` feature that should cover this functionality. All that is necessary is a republish of the existing detailed guides, like https://github.com/alphagov/whitehall/pull/2576.

PR #204, opened 2016-04-29
Inverse links for document collections

boffbowsh:
Document collections have links to documents. (Whitehall) Documents
need to know which collections they belong to. This extends similar
logic that was created for working groups and policies.

https://trello.com/c/UcqLRRSu/387-add-document-collection-documents-to-placeholder


tvararu:
👍

PR #207, opened 2016-05-03
Fill schema_name and document_type with format

boffbowsh:
As part of the split of format into schema_name and document_type, we
need to have consistent data in content store. For newer items pushed
to content store from publishing api, the schema_name and document_type
are defaulted to format if they’re not provided in the content item in
the publishing api. This backfills other content items in content store
to have the same data.

This migration takes about a minute to run on the VM. It shouldn’t
block other operations.

/cc @danielroseman @elliotcm


danielroseman:
👍  but pact tests are failing, can't see why.

PR #208, opened 2016-05-04
Split format into schema name and document type

boffbowsh:
Companion to https://github.com/alphagov/whitehall/pull/2572. Best reviewed commit by commit.

The `format` field is being replaced by `schema_name` and `document_type`. The new fields have been backfilled in #207. This PR changes content store to use the new fields to inform its behaviour around linked items and route registration. It also exposes the new fields in expanded links.

This work is required to enable the migration of HTML publications from Whitehall to the new world. 

/cc @danielroseman @gpeng

### DIFF
--- a/spec/models/content_item_spec.rb
+++ b/spec/models/content_item_spec.rb
@@ -526,14 +526,14 @@ describe ContentItem, type: :model do
       expect(@item.incoming_links("related")).to eq([@other_item])
     end
 
-    context 'with the linking_format parameter' do
+    context 'with the linking_document_type parameter' do
       before :each do
-        create(:content_item, :with_content_id, format: "a", links: { "related" => [@item.content_id] })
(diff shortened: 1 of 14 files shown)


gpeng:
Looks good to me. Probably best to wait for a second opinion from @danielroseman though...

spec/models/content_item_spec.rb
```diff
@@ -526,14 +526,14 @@
       expect(@item.incoming_links("related")).to eq([@other_item])
     end
 
-    context 'with the linking_format parameter' do
+    context 'with the linking_document_type parameter' do
       before :each do
         create(:content_item, :with_content_id, format: "a", links: { "related" => [@item.content_id] })
```
danielroseman: For clarity, maybe change `format` -> `document_type` here too.

danielroseman:
One minor comment, otherwise 👍

PR #212, opened 2016-05-10
Only attempt to resolve valid, relevant multi-type content...

steventux:
https://trello.com/c/vKokCvrb/692-content-store-needs-to-handle-attachments

Adds an additional check to ensure that multi-type content attributes
contain the correct `content_type` (eg text/html) _and_ a corresponding content
value.
Refactored to remove the need to symbolize data keys twice.

### DIFF
--- a/app/presenters/content_type_resolver.rb
+++ b/app/presenters/content_type_resolver.rb
@@ -22,23 +22,20 @@ private
   end
 
   def resolve_array(array)
-    if has_content_types?(array)
-      extract_content(array)
+    if array.all? { |v| v.is_a?(Hash) }
+      array = array.map(&:symbolize_keys)
+      content_for_type = extract_content(array)
+    end
+
+    if content_for_type
+      content_for_type[:content]
     else
       array.map { |e| resolve(e) }
     end
   end
 
-  def has_content_types?(array)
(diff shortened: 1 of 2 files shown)


app/presenters/content_type_resolver.rb
```diff
@@ -22,23 +22,17 @@ def resolve_hash(hash)
...
-  def has_content_types?(array)
-    return false unless array.all? { |v| v.is_a?(Hash) }
+      content_for_type = array.detect do |h|
+        h[:content_type] == content_type && h[:content]
+      end
 
-    hash_keys = array.flat_map(&:keys).map(&:to_sym)
-    hash_keys.include?(:content_type) && hash_keys.include?(:content)
-  end
+      return content_for_type[:content] if content_for_type
```
boffbowsh: This is returning the original value if the `content_for_type` condition doesn't pass because of the `array.detect`. It might be worth adding an explicit `return array` after this line so the behaviour is more obvious.

boffbowsh:
Tiny comment. Merge when resolved and green.

app/presenters/content_type_resolver.rb
steventux: @boffbowsh I don't think the intention was very clear in the code you've commented on, I've refactored slightly for clarity.

PR #213, opened 2016-05-17
Upgrade mongoid to 5.1

danielroseman:
Apart from just wanting to keep things up to date, this is also a very speculative fix for the memory issues we've been seeing with this app.


elliotcm:
Shall we test this as a branch on integration before merging?

Specific changes look fine.

benilovj:
The code looks legit. 👍  if it all works...

PR #214, opened 2016-05-17
Store and present ContentItem#withdrawn_notice

steventux:
Withdrawing content in a publishing application should mean the content store presents the explanation and time of the withdrawal.

Part of https://trello.com/c/78Q9mrDe/708-make-whitehall-use-unpublishing-endpoint-medium

### DIFF
--- a/spec/integration/fetching_content_item_spec.rb
+++ b/spec/integration/fetching_content_item_spec.rb
@@ -55,6 +55,7 @@ describe "Fetching content items", type: :request do
         details
         links
         expanded_links
+        withdrawn_notice
       ])
 
       expect(data).to include(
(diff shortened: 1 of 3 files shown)


spec/integration/fetching_content_item_spec.rb
```diff
@@ -260,4 +261,26 @@
...
+      )
+    end
+
+    it "displays the withdrawal explanation and time" do
+      get_content withdrawn_item
+
+      data = JSON.parse(response.body)
+
+      expect(data["withdrawn_notice"]["explanation"]).to eq("This is no longer true")
+      expect(Time.parse(data["withdrawn_notice"]["withdrawn_at"])).to eq(withdrawn_at)
```
elliotcm: We should check this against the iso8601 version to make sure that's happening.

PR #215, opened 2016-06-02
Add "links.documents" index to content_items

tvararu:
The `content-store` queries for this field when searching for detailed guides (and presumably other documents). Queries run in about 4-5 seconds, i.e.

```bash
$ mongo content_store_development
> db.content_items.find({ "links.documents": { "$in": ["5f1856a9-7631-11e4-a3cb-005056011aef"] } })
```

Adding an index appears to fix it.

The indexes will be [created automatically on deployment](https://github.com/alphagov/govuk-app-deployment/blob/master/content-store/config/deploy.rb#L21), but if you want to test this locally you can manually create them using:

```
bundle exec rake db:mongoid:create_indexes
```

### DIFF
--- a/app/models/content_item.rb
+++ b/app/models/content_item.rb
@@ -79,6 +79,11 @@ class ContentItem
   # and fetch the entire document.
   index(content_id: 1, locale: 1, format: 1, updated_at: -1, title: 1, _id: 1)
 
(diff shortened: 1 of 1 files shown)


boffbowsh:
Note this query will be going away when dependency resolution moves to publishing api from content store, at which point the index can be removed again.

tvararu:
Good point, I'll comment on that in the actual code.

boffbowsh:
Queried from here https://github.com/alphagov/content-store/blob/d066a40489c731d2dc835968750b92c8dd992f5c/app/models/content_item.rb#L122-L126

PR #217, opened 2016-06-06
Add schema validation to rendered content

elliotcm:
This is going to need a lot of schema tidyup work to make our current content validate.  I've done some (here: https://github.com/alphagov/govuk-content-schemas/pull/323) but expect lots of airbrake notifications when this goes live.

We also need to deploy the content schemas before deploying this, to get the schemas onto the content store servers (see https://github.com/alphagov/govuk-content-schemas/pull/322)

https://trello.com/c/qegWyQTI/725-content-store-validation-against-front-end-schemas


app/validators/schema_validator.rb
```diff
@@ -0,0 +1,52 @@
...
+  end
+
+private
+
+  attr_reader :rendered_content_item
+
+  def validate_schema
+    JSON::Validator.validate!(schema, rendered_content_item)
+  rescue JSON::Schema::ValidationError => error
+    Airbrake.notify_or_ignore(error, parameters: {
```
boffbowsh: The error case is likely here (not most likely but it is likely). This actually introduces errbit as a dependency for a successful call to content store. Could we make this call async, perhaps using Airbrake's built in async mode? https://github.com/airbrake/airbrake/wiki/Using-Airbrake-asynchronously

elliotcm: I've had a chat with @boffbowsh.  Airbrake's async stuff is OK-ish but we're going to look into using github scientist for this (and another thing) since we've been wanting to try it out.

elliotcm:
Closing temporarily.

PR #218, opened 2016-06-09
Add "links.working_groups" index to content_items

tvararu:
The `content-store` queries for this field when searching for guides. Queries run slowly, here's an example:

```bash
$ mongo content_store_development
> db.content_items.find({ "links.working_groups": { "$in": ["f3a2c385-2a22-47db-a532-0c343315caa3"] } })
```

Adding an index appears to fix it. Dependency resolution will also fix this, after which we may remove it.

The indexes will be [created automatically on deployment](https://github.com/alphagov/govuk-app-deployment/blob/master/content-store/config/deploy.rb#L21), but if you want to test this locally you can manually create them using:

```
bundle exec rake db:mongoid:create_indexes
```

Prior art: https://github.com/alphagov/content-store/pull/215

### DIFF
--- a/app/models/content_item.rb
+++ b/app/models/content_item.rb
@@ -84,6 +84,11 @@ class ContentItem
   # at which point this index can be removed again.
   index("links.documents" => 1)
 
+  # Add an index to speed up calls from working groups.
+  # TODO: This will go away once dependency resolution moves to publishing api from content store,
+  # at which point this index can be removed again.
+  index("links.working_groups" => 1)
+
   # We want to force the JSON representation to use "base_path" instead of
   # "_id" to prevent "_id" being exposed outside of the model.
   def as_json(options = nil)


tvararu:
Added a do not merge label as my previous PR doesn't seem to have updated the indexes on integration:

```
production:PRIMARY> db.content_items.getIndexes()
[
	{
		"v" : 1,
		"key" : {
			"_id" : 1
		},
		"ns" : "content_store_production.content_items",
		"name" : "_id_"
	},
	{
		"v" : 1,
		"key" : {
			"content_id" : 1,
			"locale" : 1,
			"format" : 1,
			"updated_at" : -1,
			"title" : 1,
			"_id" : 1
		},
		"ns" : "content_store_production.content_items",
		"name" : "content_id_1_locale_1_format_1_updated_at_-1_title_1__id_1"
	}
]
```

tvararu:
After looking into this, it seems like the previous index has rolled out on its own accord on staging. I assume the automatic creation on deployment does work, but due to the delays between merging -> deploying -> syncing mongo to `integration` it took more than 7 calendar days to actually happen for the previous index I added.

PR #219, opened 2016-06-10
Delete stale placeholder policies

benhyland:
There are a few placeholder policies which the policy-publisher
doesn't know about.

This is because those policies were renamed at some point before
the migration to publishing-api was completed.

We should delete these manually.
We already introduced another PR to deal with publishing-api.

See
https://trello.com/c/xARBi6HF
and
https://github.com/alphagov/publishing-api/pull/356


tijmenb:
Closed in favour of doing a `unpublish` API call to the publishing-api.

PR #220, opened 2016-06-17
Respond with 410 for gone items...

steventux:
Part of https://trello.com/c/od05JtVL/732-unpublishing-return-404-or-410-not-200-medium

The content store should return a response status code consistent with the router entry for this content. For 'gone' items this will be 410.


Davidslv:
👍 Looks good to me.

PR #221, opened 2016-06-22
Do not present links when they have already been expanded

dougdroper:
Links are expanded in the publishing API, so there is no need to
expand them in the content store.

### DIFF
--- a/app/models/content_item.rb
+++ b/app/models/content_item.rb
@@ -121,7 +121,14 @@ class ContentItem
 
   # Return a Hash of link types to lists of related items
   def linked_items
+    return links if already_expanded_links?
     LinkedItemsQuery.new(self).call
+  end
+
+  def already_expanded_links?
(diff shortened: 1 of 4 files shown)


app/models/content_item.rb
```diff
@@ -121,9 +121,14 @@ def gone?
 
   # Return a Hash of link types to lists of related items
   def linked_items
+    return links if already_expanded_links?
     LinkedItemsQuery.new(self).call
   end
 
+  def already_expanded_links?
+    links.to_s.match(/web_url/)
```
mobaig: maybe `links.has_key? "web_url"` instead? bit clearer what  the underlying data structure is.

dougdroper: yeah, the problem is that links have nested hashes of arbitrary types 
```javascript
{
  parent: {web_url: '/test'}
}
```

PR #222, opened 2016-06-27
Rake task to compare links with expanded links

dougdroper:
Uses HashDiff to print out the differences of links and expanded_links
to ensure there are no breaking changes

### DIFF
--- a/lib/tasks/compare.rake
+++ b/lib/tasks/compare.rake
@@ -1,0 +1,24 @@
+require 'compare_links'
+
+namespace :compare do
+  desc """
+  Compare expanded links with links for migration for a schema_name
+  rake 'compare:links[service_manual_guide]'
+  """
+  task :schema_name, [:schema_name] => :environment do |_t, args|
+    schema_name = args[:schema_name]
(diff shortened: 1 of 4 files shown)


lib/tasks/compare.rake
```diff
@@ -0,0 +1,24 @@
+require 'compare_links'
+
+namespace :compare do
+  desc """
+  Compare expanded links with links for migration for a schema_name
+  rake 'compare:links[service_manual_guide]'
+  """
+  task :schema_name, [:schema_name] => :environment do |_t, args|
+    schema_name = args[:schema_name]
+    ContentItem.where(schema_name: schema_name).each do |item|
```
danielroseman: quibble but this could be thousands of items, so probably better to use `find_each`.

danielroseman:
One minor comment otherwise 👍

lib/tasks/compare.rake
dougdroper: There isn't a `find_each` on this ContentItem. It isn't an `ActiveRecord`object

danielroseman: Ah yes, just ignore me...

PR #226, opened 2016-07-01
Add to `public_updated_at` and `updated_at` documentation

thehenster:
Hopefully this might help people to choose between the two fields in a frontend.

### DIFF
--- a/doc/content_item_fields.md
+++ b/doc/content_item_fields.md
@@ -115,8 +115,16 @@ for the Latin America and Caribbean region").
 ISO 8601 formatted timestamp. Present in all contexts.
 
 This is the update date that should be surfaced to the user. This is used for
-sorting documents by update date. It should change when there is a major
-update and should not change for a minor update.
+sorting documents by update date or to display when the document was last
+changed.
+
+If a `public_updated_at` field is included when sending the draft payload to the
+publishing api then this field will represent when the draft was saved. Publishing
+will not update it.
+
+If a `public_updated_at` field is not included when sending the draft payload
+then this field represents the time it was published for a major version, or the
+time the most recent major version was published if a minor version.
 
 ## `first_published_at`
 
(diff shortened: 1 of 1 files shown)


thehenster:
The existing document claimed the following about the `updated_at` field.. 

```
Note: This field is set by the content store whenever a item is created or modified in it.
```

.. which is a little worrying. It makes me wonder if the publishing api is setting this value and then the content store is also modifying it.

thehenster:
It looks like we intentionally want the content store to update the `updated_at` field.

If possible it would be good to remove this `updated_at` field from the public facing JSON. The name is too generic for it's meaning to be obvious. If it has to stick around in a perfect world it would be renamed to `system_updated_at` or something so at least it came with a warning flag.

danielroseman:
We don't send updated_at from publishing-api to content-store; they are completely separate.

thehenster:
Ok, I'll have a read to see what endpoints cause `updated_at` to be updated and amend it. I'm off on hols for a week so I'll pick it up then..

doc/content_item_fields.md
```diff
@@ -115,8 +115,22 @@ for the Latin America and Caribbean region").
...
-update and should not change for a minor update.
+sorting documents by update date or to display when the document was last
+changed.
+
+If a `public_updated_at` field is included when sending the draft payload to the
+publishing api then this field will represent when the draft was saved. Publishing
+will not update it.
+
+If a `public_updated_at` field is not included when sending the draft payload
+then this field represents the time it was publishing for a major version, or the
```
Davidslv: do you mean `it was published` ?

doc/content_item_fields.md
```diff
@@ -208,6 +222,12 @@ modified in it.
 It contains the timestamp at which the content was last modified in any way.
 This is suitable to be used for update versioning.
 
+When using the publishing api putting content, patching links and publishing
```
Davidslv: I find this text confusing because we are talking about requests to Publishing API and not actions.
Can you change it for something along the following lines?

```
Performing any of the following requests to publishing api will update the timestamp:

- put content
- patch links
- publish

```

doc/content_item_fields.md
thehenster: yeah - thanks!

doc/content_item_fields.md
thehenster: Thanks @Davidslv - your version was clearer but I ended up rewriting that section.

thehenster:
I've had a stab at improving my `updated_at` amendment. I tried to make it clearer that it isn't a field that is passed from the publishing-api and how it might get updated depending on whether it's a draft or live content-store, which depends on which publishing api calls are made.

I'll squash when it's good.

Davidslv:
It looks better to me 👍

PR #229, opened 2016-07-11
Sort the array of links

cbaines:
This removes noise from the diff, where its just the position of the element in
the array that differs.

The ordering of the elements is non-deterministic in the content store and
publishing API, as far as I can tell.

### DIFF
--- a/lib/compare_links.rb
+++ b/lib/compare_links.rb
@@ -38,9 +38,15 @@ private
 
   def sort(links)
     links.each_with_object({}) do |(k, v), h|
-      h[k] = v.map do |value|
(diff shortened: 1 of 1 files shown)


lib/compare_links.rb
```diff
@@ -38,9 +38,15 @@ def api_url_method
 
   def sort(links)
     links.each_with_object({}) do |(k, v), h|
-      h[k] = v.map do |value|
+      v = v.map do |value|
```
dougdroper: this is fine, although feel slightly apprehensive about clobbering the passed in variable `v`

PR #230, opened 2016-07-12
Validate against frontend schemas

steventux:
https://trello.com/c/qegWyQTI/725-content-store-validation-against-front-end-schemas

Validates the outgoing presentation of a content item against the relevant frontend schema.
Validation errors are reported via Airbrake.
This should go out with https://github.gds/gds/alphagov-deployment/pull/1204 in order to take Airbrake notification out of the request process.

### DIFF
--- a/app/validators/schema_validator.rb
+++ b/app/validators/schema_validator.rb
@@ -1,0 +1,74 @@
+require "json-schema"
+
+class SchemaValidator
+  def initialize(type:, schema_name: nil, schema: nil)
+    @type = type
+    @schema = schema
+    @schema_name = schema_name
+  end
+
+  def validate(payload)
+    @payload = payload
+
+    return true if schema_name_exception?
+
+    errors = JSON::Validator.fully_validate(
+      schema,
+      payload,
+      errors_as_objects: true,
+    )
+
+    return true if errors.empty?
+
+    errors = Hash[errors.map.with_index { |e, i| [i, present_error(e)] }]
+
+    Airbrake.notify_or_ignore(
+      {
+        error_class: "SchemaValidationError",
+        error_message: "Error validating payload against schema"
+      },
+      parameters: {
+        errors: errors,
+        message_data: payload
+      }
+    )
+    false
+  end
+
+private
+
+  attr_reader :payload, :type
+
+  def present_error(error_hash)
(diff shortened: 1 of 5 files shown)


config/initializers/sidekiq.rb
```diff
@@ -0,0 +1,21 @@
+redis_config = {
```
tijmenb: This config is not necessary if you use `govuk_sidekiq`.

steventux: Thanks @tijmenb I can see this initializer is a duplicate of same in govuk_sidekiq, will remove.

As regards to your point about the dependency between content store and redis, I agree, it feels like the tests in the async_experiments gem needs to cover the scenario of redis being unavaible and handling this gracefully. I'll put this PR on hold until I've got those changes in.

steventux:
@tijmenb : I've reverted to the simpler version of this work as there was concern that adding Redis/Sidekiq dependencies via the async_experiments gem adds complexity to the content-store.
While we can make the govuk_sidekiq gem resilient to Redis outages it's questionable whether we want the content-store accumulating these additional dependencies.

boffbowsh:
Is there a plan for surfacing the errors this raises? Will it be actively monitored?

steventux:
@boffbowsh the current plan is to see how much noise this creates in Errbit on integration as an initial step to understanding which formats might be a problem. If it proves too noisy I think we'll need to discuss alternatives to Airbrake, maybe stats for failing formats.

boffbowsh:
👍

app/validators/schema_validator.rb
```diff
@@ -0,0 +1,74 @@
...
+
+    errors = JSON::Validator.fully_validate(
+      schema,
+      payload,
+      errors_as_objects: true,
+    )
+
+    return true if errors.empty?
+
+    errors = errors.map { |e| present_error(e) }
```
danielroseman: I made a tweak to this and the present_error method itself in https://github.com/alphagov/publishing-api/pull/417/commits/270d6b6f8cbce9cd0f010058e9fc1c0bf87b11b5 because Airbrake/Errbit doesn't understand arrays, so shows the list of errors as a hard-to-read string.

steventux:
@danielroseman updated/rebased

PR #231, opened 2016-07-13
Use expanded links instead of expanding the links in the content_store

dougdroper:
Travel advice document_types still need to go through the old flow

### DIFF
--- a/app/presenters/content_item_presenter.rb
+++ b/app/presenters/content_item_presenter.rb
@@ -31,29 +31,31 @@ class ContentItemPresenter
   end
 
   def as_json(options = nil)
-    @item.as_json(options).slice(*PUBLIC_ATTRIBUTES).merge(
+    item.as_json(options).slice(*PUBLIC_ATTRIBUTES).merge(
       "links" => links,
-      "description" => RESOLVER.resolve(@item.description),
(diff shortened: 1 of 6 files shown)


app/presenters/content_item_presenter.rb
```diff
@@ -42,6 +42,7 @@ def as_json(options = nil)
 private
 
   def links
+    return @item.expanded_links unless document_type_expanded?
```
danielroseman: Should we remove `expanded_links` from the PUBLIC_ATTRIBUTES above?

dougdroper: Yes definitely, updated.

app/presenters/content_item_presenter.rb
```diff
@@ -52,4 +52,8 @@ def links
   def present_linked_item(linked_item)
     LinkedItemPresenter.new(linked_item, @api_url_method).present
   end
+
+  def document_type_expanded?
+    @item.document_type =~ /travel_advice/
```
gpeng: `attr_reader :item, :api_url_method` would avoid accessing the instance variables directly

dougdroper: thanks for the handy tip 👍  updated

PR #235, opened 2016-08-10
Remove expanded links

dougdroper:
The publishing API is now doing most of the link expansion/dependency
resolution. Passthrough hashes are still being handled here,
but will also be removed when they are removed form the publishing
API/travel advice.

This also removes the compare link rake task
Links are no longer expanded in the content store, so this is
unnecessary.

dependent on https://trello.com/c/yNL1Lziv/727-prepare-formats-to-switch-to-expanded-links

https://trello.com/c/Ndqxkagy/768-post-dependency-resolution-test-and-then-remove-link-expansion-in-content-store-medium


tijmenb:
🎉

app/presenters/content_item_presenter.rb
```diff
@@ -44,14 +44,28 @@ def as_json(options = nil)
...
-        items[link_type] = linked_items.map { |i| present_linked_item(i) }
-      end
+    passthrough.merge(available_translations: available_translations)
+  end
+
+  def passthrough
+    item.links.each_with_object({}) do |(link_type, passthrough_hash), result|
+      passthrough, ids = passthrough_hash.partition { |type| type.is_a?(Hash) }
+      hashes = passthrough.map { |attributes| present(ContentItem.new(attributes)) }
+      content_items = ids.map { |id| present(ContentItem.find_by(content_id: id)) }
```
kevindew: Not sure if intended or not as ActiveRecord & Mongoid differ here but `find_by` will raise an exception if an item isn't found here - I'm not sure how likely that scenario is.
Also could be more performant and easier to read if this was `content_items = ContentItem.where(:id.in => ids).map ... `

kevindew:
Aside from the one comment this looks good, not a single conditional outside of a guard condition 💥

app/presenters/content_item_presenter.rb
dougdroper: Yeah, very unlikely in this case, but also no need to be that aggressive 👍 
I ended up taking the old query, since it did a bit more filtering (also unnecessary) but good to keep a reference too until this code is completely removed.

app/presenters/content_item_presenter.rb
```diff
@@ -51,11 +51,18 @@ def passthrough
...
       passthrough, ids = passthrough_hash.partition { |type| type.is_a?(Hash) }
       hashes = passthrough.map { |attributes| present(ContentItem.new(attributes)) }
-      content_items = ids.map { |id| present(ContentItem.find_by(content_id: id)) }
+      content_items = content_items_for(ids).map { |content_item| present(content_item) }
       result[link_type] = hashes + content_items
     end
   end
 
+  def content_items_for(content_ids)
+    renderable_items_by_id = ContentItem
```
kevindew: You just need to kill this assignment

dougdroper: cool, done

PR #236, opened 2016-08-11
Allow `gone` items to be returned with 200

gpeng:
Currently sending a `gone` item through the `/unpublish` endpoint of Publishing API always creates a `gone` route in the router and causes Content Store to also return a 410 status.

Due to the vagueries of Whitehall 'unpublishing' we need a way to unpublish but still render an explanation and optional alternative url. This can't be done with a 410 route in the router.

We need to address this properly but for now this change should allow the item to be rendered correctly and fix the issues we have in production.

### DIFF
--- a/app/models/content_item.rb
+++ b/app/models/content_item.rb
@@ -121,7 +121,13 @@ class ContentItem
   end
 
   def gone?
-    self.schema_name == "gone"
+    #we've overloaded gone a bit by adding explanation and alternative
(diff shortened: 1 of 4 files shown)


boffbowsh:
I think this also needs a spec that if you request a `gone` content item with details, it actually returns the details.

gpeng:
@boffbowsh added in 09a0970

PR #239, opened 2016-09-09
Remove workarounds for passthrough hash

dougdroper:
Now that multipage-frontend no longer requires any special passthrough
hash this can be removed

- [x] https://github.com/alphagov/multipage-frontend/pull/44 to be merged and deployed


kevindew:
Corr look at that add to remove ratio ✂️

PR #240, opened 2016-09-20
Fix a test in the fetching linked items spec

cbaines:
Initially it appeared that my local timezone was leaking in to the test,
and causing the public_updated_at values to differ by an hour in the
actual and expected response. I tried quite a few different ways of
resolving this, and some worked, but led to the times sometimes being a
millisecond apart.

Changing to using travel_to (instead of Timecop) and comparing against
the values from the created objects was the only way I could get the
test to pass reliably.

I was expecting that just comparing against the values on the objects
would resolve any issues, but, the rendering of the time differed. I
think the value of public_updated_at used to construct the request body
in the presenter may come directly from the factory, but in the test,
the value comes from the database, which is leading to a disparity.

### DIFF
--- a/spec/integration/fetching_linked_items_spec.rb
+++ b/spec/integration/fetching_linked_items_spec.rb
@@ -1,46 +1,47 @@
 require 'rails_helper'
+include ActiveSupport::Testing::TimeHelpers
 
 describe "Fetching linked items", type: :request do
   describe "GET /incoming-links/:base_path_without_root" do
     it "returns the items linked to an item" do
-      Timecop.freeze do
-        item = create(:content_item, :with_content_id, document_type: "travel_advice")
-        create(:content_item, content_id: 'ID-1', base_path: '/a', title: "A", links: { "parent" => [item.content_id] })
-        create(:content_item, content_id: 'ID-2', base_path: '/b', title: "B", links: { "parent" => [item.content_id] })
+      item = create(:content_item, :with_content_id, document_type: "travel_advice")
+      id_1 = create(:content_item, content_id: 'ID-1', base_path: '/a', title: "A", links: { "parent" => [item.content_id] })
+      id_1.reload
+      id_2 = create(:content_item, content_id: 'ID-2', base_path: '/b', title: "B", links: { "parent" => [item.content_id] })
+      id_2.reload
 
-        get "/incoming-links#{item.base_path}?types[]=parent&types[]=topics"
+      get "/incoming-links#{item.base_path}?types[]=parent&types[]=topics"
 
-        expect(parsed_response["parent"]).to eql([
-          {
-            "content_id" => "ID-1",
-            "title" => "A",
-            "base_path" => "/a",
-            "description" => nil,
-            "api_url" => "http://www.example.com/content/a",
-            "web_url" => "https://www.test.gov.uk/a",
-            "locale" => "en",
-            "public_updated_at" => Time.now.as_json,
-            "schema_name" => "answer",
-            "document_type" => "answer",
-            "links" => { "parent" => [item.content_id] },
-          },
-          {
-            "content_id" => "ID-2",
-            "title" => "B",
-            "base_path" => "/b",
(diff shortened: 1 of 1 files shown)


spec/integration/fetching_linked_items_spec.rb
```diff
@@ -1,12 +1,13 @@
 require 'rails_helper'
+include ActiveSupport::Testing::TimeHelpers
 
 describe "Fetching linked items", type: :request do
   describe "GET /incoming-links/:base_path_without_root" do
     it "returns the items linked to an item" do
-      Timecop.freeze do
+      travel_to Time.now do
```
carlosmartinez: Did you try calling `Timecop.freeze` in a `before` block? I suspect that might work (although it doesn't really matter, since you've got it working another way)

cbaines: I didn't at the time, but I did just now, and it looks like using a before block is no different (I still see a different timezone).

My suspicion is that the "wrong" value I am seeing is coming from Mongoid, which does use ActiveSupport for some time related things, so maybe it has better support for manipulating Mongoid than Timecop...

kevindew:
Did you try replacing `Time.now` with `Time.now.utc` ?

cbaines:
> Did you try replacing Time.now with Time.now.utc ?

I looked at doing that, but the content item factory uses Time.now, so I didn't want to make the test and the factory differ.

carlosmartinez:
There's a good case for making both of them `Time.now.utc`

kevindew:
@cbaines are you planning on making any changes here? I'd have thought if you want to keep it as comparing times to the model that you wouldn't need the `travel_to Time.now do` block

cbaines:
> @cbaines are you planning on making any changes here? I'd have thought if you want to keep it as comparing times to the model that you wouldn't need the travel_to Time.now do block

Not at the moment. For me, when I remove that block, the test fails non-deterministically (with the expected and actual public_updated_at time differing by 1ms). I'm not particularly interested in working out why this is (at least at the moment), so I'd prefer to just keep the travel_to block, as it appears to prevent this from failing.

I started to dig in to the why before I'd found travel_to, and my least worst theory is that it has something to do with the way the time data is passing through MongoDB, and also, possibly not passing through MongoDB the first time the factory is used...

kevindew:
Sure, I had a quick poke myself as it seemed pretty well that the time was changing against what it's initialised at and it looks like the non-deterministic aspect is caused by the precision with which mongo stores the date time (to 3 digits in microseconds). 

`travel_to` masks the effect of that by setting the [microseconds to 0](https://github.com/rails/rails/blob/master/activesupport/lib/active_support/testing/time_helpers.rb#L134)

Should work fine if you reload the objects once creating them to get the db representation and then you should be able to drop the `travel` to

cbaines:
> Sure, I had a quick poke myself as it seemed pretty well that the time was changing against what it's initialised at and it looks like the non-deterministic aspect is caused by the precision with which mongo stores the date time (to 3 digits in microseconds).
> 
> travel_to masks the effect of that by setting the microseconds to 0
> 
> Should work fine if you reload the objects once creating them to get the db representation and then you should be able to drop the travel_to

Thanks for investigating further Kevin, I've tested this, and it seems to work, but I don't really understand why? Is there any documentation on what the reload method does (the source is less than clear https://github.com/thoughtbot/factory_girl/blob/master/lib/factory_girl/reload.rb ).

cbaines:
> Thanks for investigating further Kevin, I've tested this, and it seems to work, but I don't really understand why? Is there any documentation on what the reload method does (the source is less than clear https://github.com/thoughtbot/factory_girl/blob/master/lib/factory_girl/reload.rb ).

Talked offline, I didn't realise that the thing returned from the factory was an actual model, and not something from FactoryGirl, so the reload method was/is a mongoid thing.

cbaines:
Branch now updated to use .reload and not use travel_to.

kevindew:
awesome 👍

PR #241, opened 2016-09-20
Log Publishing-Source-Dependent-Content-Id header

cbaines:
The presence of a value for this can be used to distinguish requests
that are a product of dependency resolution in the Publishing API, and
those that are not.

The value indicates the content id of the source of the dependency
resolution activity.


cbaines:
Now updated per https://github.com/alphagov/publishing-api/pull/517#issuecomment-248924432 .

PR #245, opened 2016-10-05
Plek based link urls

kevindew:
We have a problem that due to data syncing the values of `web_url` and `api_url` in the expanded links of a content item reference the production environment. This causes problems for links in the staging and integration environment.

The ideal way forward is for rendering applications to no longer use urls and instead use the absolute paths of base_path and api_path, therefore we plan to deprecate these fields from the content store.

In the meantime though this dynamically sets web_url and api_url based off Plek to be relevant to the environment the content store is running within.

### DIFF
--- a/app/presenters/content_item_presenter.rb
+++ b/app/presenters/content_item_presenter.rb
@@ -43,6 +43,6 @@ private
   attr_reader :item, :api_url_method
 
   def links
-    item.expanded_links
+    ExpandedLinksPresenter.new(item.expanded_links).present
   end
 end
--- a/app/presenters/expanded_links_presenter.rb
+++ b/app/presenters/expanded_links_presenter.rb
@@ -1,0 +1,37 @@
+class ExpandedLinksPresenter
+  attr_reader :expanded_links
+
+  def initialize(expanded_links)
+    @expanded_links = expanded_links
+  end
+
+  def present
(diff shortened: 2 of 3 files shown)


app/models/content_item.rb
```diff
@@ -2,7 +2,7 @@ class ContentItem
   include Mongoid::Document
   include Mongoid::Timestamps::Updated
 
-  NON_RENDERABLE_FORMATS = %w(redirect gone)
+  NON_RENDERABLE_FORMATS = %w(redirect gone).freeze
```
dougdroper: is this used anywhere?

kevindew: Yup - at [line 70](https://github.com/alphagov/content-store/blob/506c2c802a6e88c766d8e352170d290ea64c99c0/app/models/content_item.rb#L70)

dougdroper: And is line 70 used anywhere?

kevindew: Yup - though I've realised it's just in a (2 year old) [migration](https://github.com/alphagov/content-store/blob/master/db/migrate/20141126123135_add_locale_to_existing_content.rb)

So I imagine it is safe to kill.

kevindew: I've updated this PR not to include this. Will fire in another with that constant removed.

PR #247, opened 2016-10-17
Remove force_http when registering a backend

alexmuller:
@jennyd pointed this out last week - I think this exists from a time when the router couldn't handle HTTPS backends. It can now because we do HTTPS to Trade Tariff and Licensing.

https://github.com/alphagov/router/commit/7e72e8b073bedb05866a0bd68e76ffdac2af7f1e

This will cause new registrations to use the HTTPS backend which will encrypt traffic in our network.

### DIFF
--- a/app/models/route_set.rb
+++ b/app/models/route_set.rb
@@ -84,7 +84,7 @@ class RouteSet < OpenStruct
 private
 
   def register_rendering_app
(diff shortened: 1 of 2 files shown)


tijmenb:
Perhaps after deploying this we should change all current entries into `https` manually so that we know the effect instantly, instead of waiting for a republish of all apps.

cbaines:
:+1: I also like the idea of proactively updating the existing backends.

PR #250, opened 2016-11-29
Don't expose `format`

tijmenb:
This stops content store from exposing `format` to users. This attribute is deprecated in favour of `schema_name` and `document_type`.

Removing the attribute from the database is out of scope for this PR.

https://github.com/alphagov/govuk-content-schemas/pull/444 removes the attribute from the frontend schemas.

https://trello.com/c/C8A9bnuO

cc @danielroseman is this sensible?

### DIFF
--- a/app/presenters/content_item_presenter.rb
+++ b/app/presenters/content_item_presenter.rb
@@ -12,7 +12,6 @@ class ContentItemPresenter
     content_id
     document_type
     first_published_at
-    format
     locale
     need_ids
     phase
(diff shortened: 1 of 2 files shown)


tijmenb:
@danielroseman what do you think?

danielroseman:
I guess this is OK. There's a bunch of code in government-frontend that refers to a format attribute, but it looks like it actually gets that data from schema_name anyway.

tijmenb:
📿

PR #251, opened 2016-11-29
Add a little bit of schema validation for output

tijmenb:
This adds a test that the output of `ContentItemPresenter` is valid against the `generic` schema. This will ensure that the schemas and this app will be kept in sync.

https://trello.com/c/C8A9bnuO


rubenarakelyan:
🎉

PR #253, opened 2016-12-01
Remove deleted content items

gpeng:
Adds a migration to delete a number of content items. The items are `HtmlAttachment`s that have been published from Whitehall but are in fact deleted and should not be in the content store. They all currently 404 so these items need to be removed before we switch the rendering of `HtmlAttachment` content to content-store -> government-frontend.


steventux:
:+1: Merge when you're ready

PR #258, opened 2016-12-09
Add Jenkinsfile and remove gh-status calls.

danielroseman:
Before merging, we need to run the following on the Jenkins script console:

    def job = Jenkins.instance.getItemByFullName("content-store/master")
    job.nextBuildNumber = 598
    job.save()


kevindew:
I don't think this will blow up on a linting error, though it doesn't look like that was handled before?

Changes look fine though

PR #267, opened 2017-03-03
Always register routes, even if the route set is unchanged

cbaines:
Conditionally registering the routes could lead to routes not being
registered entirely for some content. Consider the following scenario:

 - Start with an empty content store
 - update request made
 - No previous item
 - Item saved to the database
 - Route registration fails
 - Content Store returns error status to client

 - update request retried by client
 - previous_item exists
 - route registration skips, as route_set matches that of the previous
   item
 - Content Store returns success status to client

Even though the 2nd request is "successful", the routes have not been
registered. This can also happen if the item exists in the content
store, but the update has different routes.

Always registering the route seems preferable, as the code is simpler
compared to manually rolling back the creation or update of the local
database, however implementing rollback manually in the code would be
a possible alternative, and this change will result in some additional
redundant requests to the router-api.

[Trello](https://trello.com/c/un4CpSs4/670-figure-out-why-check-tenant-right-to-rent-documents-and-postgraduate-loan-are-not-previewable)

### DIFF
--- a/app/models/content_item.rb
+++ b/app/models/content_item.rb
@@ -14,7 +14,7 @@ class ContentItem
     item.assign_attributes(attributes)
 
     if item.upsert
-      item.register_routes(previous_item: previous_item)
+      item.register_routes
     else
       result = false
     end
@@ -138,7 +138,7 @@ class ContentItem
     fact_check_ids.include?(fact_check_id)
   end
 
-  def register_routes(previous_item: nil)
+  def register_routes
     return if self.schema_name.start_with?("placeholder")
     self.route_set.register!
   end
--- a/app/models/content_item.rb
+++ b/app/models/content_item.rb
@@ -140,7 +140,6 @@ class ContentItem
 
   def register_routes(previous_item: nil)
     return if self.schema_name.start_with?("placeholder")
-    return if previous_item && previous_item.route_set == self.route_set
     self.route_set.register!
   end
 
--- a/spec/integration/submitting_content_item_spec.rb
+++ b/spec/integration/submitting_content_item_spec.rb
@@ -136,12 +136,7 @@ describe "content item write API", type: :request do
       expect(@item.details).to eq("body" => "<p>Some body text</p>\n")
     end
 
-    it "does not register routes when they haven't changed" do
-      put_json "/content/vat-rates", @data
-      refute_routes_registered("frontend", [['/vat-rates', 'exact']])
-    end
-
-    it "registers routes for the content item when they have changed" do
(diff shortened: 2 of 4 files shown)


kevindew:
Makes sense, be good to kill off the `previous_item` argument too.

Do you know if we're at much risk of this clobbering things in Router API? I'm hoping not as it makes things rather confusing.

@thomasleese would the script you put together for router/content store consistency be useful here?

cbaines:
> Makes sense, be good to kill off the previous_item argument too.

Yep, done :)

> Do you know if we're at much risk of this clobbering things in Router API? I'm hoping not as it makes things rather confusing.

I'm unsure what you mean, but I think this should only help to ensure that the router-api is more consistent with the routing data in the content stores.

tijmenb:
I've got 2 issues with this:

- There might be routes overridden by `router-data`. These are now "safe" because the routes don't change anymore, so will never be sent to the router. It's likely that either a mass-re-representation or dependency resolution will trigger this.
- In case of a mass-republishing this could trigger 350K needless requests to router-api. Not sure how well the router-api can handle that, and if this will delay publishing.

If we do want to do this, we have to make 100% sure that router-api and content-store are in sync, and do some load & performance testing.

cbaines:
> I've got 2 issues with this:
> 
>     There might be routes overridden by router-data. These are now "safe" because the routes don't change anymore, so will never be sent to the router. It's likely that either a mass-re-representation or dependency resolution will trigger this.

Hmm, yeah, I think some analysis of the data is the only way to dig in to this.

>     In case of a mass-republishing this could trigger 350K needless requests to router-api. Not sure how well the router-api can handle that, and if this will delay publishing.

I hope that it would handle it, however I would guess that for large operations this may cause a slight slowdown.

> If we do want to do this, we have to make 100% sure that router-api and content-store are in sync, and do some load & performance testing.

I've been thinking on how to do the consistency checking, and I think the Publishing Platform team did some work on this?

thomasleese:
There [is a script which checks for consistent routes](https://github.com/alphagov/content-store/blob/master/app/models/content_consistency_checker.rb) between the `content-store` and the `router-api` which would help here. Nothing has been done about setting up automatic alerts on this happening yet, although [there is a card on our board](https://trello.com/c/0uqM1YHL/434-create-alert-for-new-inconsistent-documents).

kevindew:
I'm going to close this until we get to look at: https://trello.com/c/1M8bgdQ0/452-ensure-the-content-store-reliably-updates-the-router-api just so this doesn't linger around.

PR #268, opened 2017-03-06
Add support for supertypes to content item

tijmenb:
These were added in https://github.com/alphagov/publishing-api/pull/805.

Somehow our testing system didn't pick up that content store would reject these requests. I'm planning to look into getting that fixed.

https://trello.com/c/07lAcDtC


suzannehamilton:
🚀

PR #269, opened 2017-03-08
Handling requests for content within routes

kevindew:
This changes the GET content and GET publish-intent endpoints so that they can accept a request for content which is within the routes of an item. For both cases if the path matches a route and not a base_path a 303 redirect will be sent to the route with a base_path.

When this is deployed we need to update indexes for mongoid.

Unfortunately a bunch of dependencies were needed to be updated to get this all working and they're somewhat tangled in this.


app/models/publish_intent.rb
```diff
@@ -58,6 +58,12 @@ def self.cleanup_expired
     where(:publish_time.lt => PUBLISH_TIME_LEEWAY.ago).delete_all
   end
 
+  def base_path_without_root
```
thomasleese: This could be changed to `base_path.sub(%r{^/}, "") if base_path`.

lib/tasks/pact.rake
```diff
@@ -17,7 +17,7 @@ unless Rails.env.production?
     end
   end
 
-  task default: "pact:verify"
+  # task default: "pact:verify"
```
thomasleese: Will we add a card or put a FIXME comment to avoid forgetting about this?

kevindew: good point, I've included it in acceptance criteria of https://trello.com/c/qjoezykg/901-reinstate-the-pact-broker-2

app/models/publish_intent.rb
kevindew: yes that'd be better, thanks - (was blindly c+ping methods). @dougdroper would not be happy with the `return nil` there

thomasleese: Indeed. I was making sure we still have the @dougdroper spirit on the team

app/controllers/publish_intents_controller.rb
```diff
@@ -2,19 +2,23 @@ class PublishIntentsController < ApplicationController
...
 
   def update
     result, intent = PublishIntent.create_or_update(encoded_base_path, @request_data)
 
-    if result
-      status = (result == :created ? :created : :ok)
-    else
-      status = :unprocessable_entity
-    end
+    status = if result
```
kevindew: it does indeed, good suggestion

kevindew:
@thomasleese comments addressed, thanks

PR #271, opened 2017-03-16
Rename fact_check_ids to auth_bypass_ids

brenetic:
Content needs to be previewable by users without signon accounts in most stages of a publishers workflow, not just during fact check.

https://trello.com/c/7oRBYwnn/690-rename-fact-check-id-to-auth-bypass-id

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -21,7 +21,7 @@ gem 'hashdiff', require: false
 if ENV['GDS_API_ADAPTERS_DEV']
   gem 'gds-api-adapters', path: '../gds-api-adapters'
 else
-  gem 'gds-api-adapters', '36.0.1'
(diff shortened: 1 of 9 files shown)


Gemfile
```diff
@@ -21,7 +21,7 @@ gem 'hashdiff', require: false
 if ENV['GDS_API_ADAPTERS_DEV']
   gem 'gds-api-adapters', path: '../gds-api-adapters'
 else
-  gem 'gds-api-adapters', '36.0.1'
+  gem 'gds-api-adapters', "~> 41.0.0"
```
kevindew: Could we loosen this down to 41.0 please?

brenetic: Loosened :)

PR #274, opened 2017-03-21
Send an Airbrake error when inconsistent content items are found

thomasleese:
This will allow the task to run automatically and send us an alert when inconsistent documents are found.

### DIFF
--- a/lib/tasks/check_content_consistency.rake
+++ b/lib/tasks/check_content_consistency.rake
@@ -1,6 +1,6 @@
 namespace :check_content_consistency do
-  def check_content(base_path)
-    checker = ContentConsistencyChecker.new(base_path)
+  def check_content(base_path, ignore_recent = false)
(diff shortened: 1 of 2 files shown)


lib/tasks/check_content_consistency.rake
```diff
@@ -1,11 +1,16 @@
...
-    checker = ContentConsistencyChecker.new(base_path)
+  def check_content(base_path, ignore_recent = false)
+    checker = ContentConsistencyChecker.new(base_path, ignore_recent)
     errors = checker.call
 
     if errors.any?
-      puts "#{base_path} 😱"
-      puts errors
+      Airbrake.notify(
+        "Found an inconsistent content item: #{base_path} 😱",
```
tijmenb: Same as https://github.com/alphagov/publishing-api/pull/852#discussion_r107231488. This might overload Errbit's email system.

PR #275, opened 2017-03-23
Restructure how consistency checker works

thomasleese:
It now takes a dump of the routes from the `router-api`, combined with the data from the `router-data` repository, to compare the routes. This means there is a dramatic decrease in how long it takes to complete.


lib/route_consistency_checker.rb
```diff
@@ -0,0 +1,145 @@
...
+private
+
+  def content_items_to_check
+    ContentItem
+      .where(:content_id.nin => ["", nil])
+      .where(:schema_name.not => /^placeholder/)
+  end
+
+  def load_routes(filename)
+    routes = {}
```
kevindew: maybe don't need to define this if we can do an each_with_object in the block?

lib/route_consistency_checker.rb
```diff
@@ -0,0 +1,145 @@
...
+
+      @errors[route["incoming_path"]] << "No content item available."
+    end
+  end
+
+private
+
+  def content_items_to_check
+    ContentItem
+      .where(:content_id.nin => ["", nil])
```
kevindew: why do we do this? (when your memory jogs)

lib/route_consistency_checker.rb
```diff
@@ -0,0 +1,145 @@
...
+  def check_redirect(content_item, redirect)
+    path = redirect[:path]
+
+    res = get_route(path)
+    return unless res
+
+    return if content_item.updated_at > res["updated_at"]
+
+    if res["handler"] != "redirect"
+      @errors[content_item.base_path] << "Handler is not a redirect for " \
```
kevindew: maybe have a method that we use to register an error?

lib/route_consistency_checker.rb
```diff
@@ -0,0 +1,145 @@
...
+  def check_routes
+    @remaining_routes.each do |path|
+      next if @router_data.include?(path)
+
+      route = @routes.fetch(path)
+      next if route["disabled"]
+      next if route["handler"] == "gone"
+      next if route["handler"] == "redirect" && !route["backend_id"].present?
+
+      content_item = ContentItem.find_by_path(route["incoming_path"])
```
kevindew: we want to look up the content item by an exact route match here - I don't think we're worried about base_path

lib/route_consistency_checker.rb
```diff
@@ -0,0 +1,145 @@
...
+
+    return if content_item.updated_at > res["updated_at"]
+
+    if res["handler"] != "redirect"
+      @errors[content_item.base_path] << "Handler is not a redirect for " \
+                                         "#{path}."
+    end
+
+    if res["redirect_to"] != redirect[:destination]
+      @errors[content_item.base_path] << "Route destination " \
```
kevindew: should be registering the error against the route I hear I believe rather than content_item base_path?

lib/route_consistency_checker.rb
```diff
@@ -0,0 +1,145 @@
...
+
+  def initialize(routes, router_data)
+    @routes = load_routes(routes)
+    @remaining_routes = Set.new(@routes.keys)
+    @errors = Hash.new { |hash, key| hash[key] = [] }
+    @router_data = load_router_data(router_data)
+  end
+
+  def check_content
+    content_items_to_check.each do |content_item|
```
kevindew: maybe see if we can memoize things that are checked here so they can be subtracted from routes to check? to try make remaining routes usage more explicit

lib/route_consistency_checker.rb
```diff
@@ -0,0 +1,145 @@
...
+      end
+    end
+
+    routes
+  end
+
+  def load_router_data(path)
+    routes = Set.new
+
+    Dir.glob(path + "/data/*.csv").each do |csv_path|
```
kevindew: maybe this can be an input?

lib/route_consistency_checker.rb
```diff
@@ -0,0 +1,145 @@
+require "set"
+
+require "gds_api/router"
+
+class RouteConsistencyChecker
+  attr_reader :errors
+
+  def initialize(routes, router_data)
```
kevindew: on longer discussion - we thought about making it so this only accepts data and that we have a specific class that converts the CSV into open structs

lib/route_consistency_checker.rb
```diff
@@ -0,0 +1,135 @@
...
+  attr_reader :routes, :router_data, :checked_routes
+  attr_writer :errors
+
+  def content_items_to_check
+    ContentItem
+      .where(:content_id.nin => ["", nil])
+      .where(:schema_name.not => /^placeholder/)
+  end
+
+  def remaining_routes
```
kevindew: this might be more idiomatic named as `unchecked_routes` wdyt?

lib/route_consistency_checker.rb
```diff
@@ -0,0 +1,135 @@
...
+
+  def find_content_item(path)
+    # We cannot use an 'or' query here as it seems to be slower than doing this
+    # irb(main):019:0> time { ContentItem.where(routes: { "$elemMatch" => { path: path } }).first }
+    #   0.000000   0.000000   0.000000 (  0.006512)
+    # irb(main):020:0> time { ContentItem.where(redirects: { "$elemMatch" => { path: path } }).first }
+    #   0.010000   0.000000   0.010000 (  0.004688)
+    # irb(main):021:0> time { ContentItem.or({ redirects: { "$elemMatch" => { path: path } } }, { routes: { "$elemMatch" => { path: path } } }).first }
+    #   0.000000   0.000000   0.000000 ( 11.314430)
+    find_content_item_by_route(path) || find_content_item_by_redirect(path)
```
kevindew: is it a concern of ours (ie within scope of this) where there could be either multiples of these or both of these? - just thinking that we only do something on the first thing we found whereas there is capacity for this to be multiple.

Also I'd guess (but don't know) that we know from the router data whether this is a redirect or not so might be able to limit the scope of this or based on the data?

lib/route_consistency_checker.rb
```diff
@@ -0,0 +1,135 @@
...
+  end
+
+  def check_redirect(content_item, redirect)
+    path = redirect[:path]
+
+    res = get_route(path)
+    return unless res
```
kevindew: is res short for result? probably better to have a full name

thomasleese:
@kevindew This is ready for re-review now.

lib/route_consistency_checker.rb
```diff
@@ -0,0 +1,143 @@
...
+  end
+
+  def expected_handler(content_item)
+    if content_item.gone?
+      "gone"
+    elsif content_item.redirect?
+      "redirect"
+    else
+      "backend"
+    end
```
kevindew: might be cleaner as
return "gone" if content_item.gone?
return "redirect" if content_item.redirect?
"backend"

lib/route_consistency_checker.rb
```diff
@@ -0,0 +1,143 @@
...
+  end
+
+  def check_route(content_item, route)
+    path = route[:path]
+
+    result = get_route(path)
+    return unless result
+
+    return if content_item.updated_at > result.updated_at
```
kevindew: might want to combine these two returns?

lib/route_consistency_checker.rb
```diff
@@ -0,0 +1,143 @@
...
+      nil
+    end
+  end
+
+  def find_content_item_for_field(path, kind)
+    items = ContentItem.where("#{kind}s.path" => path).entries
+    return nil if items.empty?
+    return items.first if items.length == 1
+    errors[path] << "Multiple content items returned for #{kind}."
+    nil
```
kevindew: I feel like this would read better in a if else block - to me it feels like it's supposed to return nil whereas you want it to look more like multiple items is an exceptional circumstance

thomasleese:
@kevindew Thanks, those should be addressed now.

PR #276, opened 2017-03-23
Add a content dumper

thomasleese:
This will be used by the `publishing-api` consistency checker.

### DIFF
--- a/lib/content_dumper.rb
+++ b/lib/content_dumper.rb
@@ -3,6 +3,7 @@ require "zlib"
 
 class ContentDumper
   FIELDS = %w(base_path content_id locale document_type schema_name rendering_app publishing_app updated_at).freeze
+  HASH_FIELDS = %w(details expanded_links routes redirects).freeze
 
   def initialize(filename)
     @filename = filename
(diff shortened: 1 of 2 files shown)


lib/content_dumper.rb
```diff
@@ -0,0 +1,51 @@
+require "csv"
+require "zlib"
+
+class ContentDumper
+  FIELDS = %w(base_path content_id locale document_type schema_name rendering_app publishing_app updated_at).freeze
+  HASH_FIELDS = %w(details expanded_links).freeze
```
kevindew: we probably also want routes and redirects here - unless I've missed something

lib/content_dumper.rb
```diff
@@ -0,0 +1,51 @@
...
+  end
+
+  def csv_fields(item)
+    FIELDS.map { |field| item.send(field) } + HASH_FIELDS.map { |field| hash_field(item.send(field)) }
+  end
+
+  def hash_field(object)
+    Digest::MD5.hexdigest(
+      JSON.generate(
+        deep_sort_hash(object)
```
kevindew: looking at Pub API these are sorted: https://github.com/alphagov/publishing-api/blob/master/app/queries/edition_links_from.rb#L29 so I think we can probably simplify this down to just running the JSON.generate on it.

Might worth adding a note about how the more expected `.to_json` wasn't used as we get more consistent results with JSON.generate

Also did you get a speed boost with MD5? I was just thinking that for someone reading the use of that over SHA1 for instance might cause alarm bells so if we're getting extra performance that should be noted in commit or even a comment and if it's negligible might want to use a SHA one?

kevindew: Just to add, if you've seen problems with ordering that this deep sort fixes might want to ignore my suggestion. I'm working on assumption that this is pre-empting a problem rather than fixing one.

PR #282, opened 2017-04-11
Support passing segments mode to router api

cbaines:
https://trello.com/c/lK2SWzGS/39-fix-content-store-publishing-api-bug-as-segment-information-is-not-being-passed


pmanrubia:
LGTM

PR #286, opened 2017-04-28
Add register_backends rake task

cbaines:
This provides a convinient way to set backends in the Router API, for
all the content in the Content Store.

This can be useful for a few reasons:
 - You have changed the configuration of the
content store with respect to the backends it generates, and want to
update the Router API.
 - The data in the Router API doesn't match, as you have loaded the
database from another a deployment of these applications with
different configuration.

### DIFF
--- a/lib/tasks/register_backends.rake
+++ b/lib/tasks/register_backends.rake
@@ -1,0 +1,11 @@
+desc "
+  Register backends with the router-api for all rendering apps in the
+  database.
+"
+task register_backends: :environment do
+  ContentItem.distinct(:rendering_app).compact.each do |rendering_app|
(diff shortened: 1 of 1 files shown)


lib/tasks/register_backends.rake
```diff
@@ -0,0 +1,11 @@
+desc "
+  Register backends with the router-api for all rendering apps in the
+  database.
+"
+task register_backends: :environment do
+  ContentItem.all.distinct(:rendering_app).compact.each do |rendering_app|
```
kevindew: super minor but you don't need `.all` here

cbaines: Thanks, I've removed `.all` now :)

PR #287, opened 2017-04-28
Remove transmitted_at support

emmabeynon:
https://trello.com/c/fGsTnEVZ/925-remove-transmitted-at-support-from-content-store

### DIFF
--- a/app/models/content_item.rb
+++ b/app/models/content_item.rb
@@ -6,7 +6,8 @@ class ContentItem
     previous_item = ContentItem.where(base_path: base_path).first
     lock = UpdateLock.new(previous_item)
 
-    lock.check_availability!(attributes)
+    payload_version = attributes["payload_version"]
+    lock.check_availability!(payload_version)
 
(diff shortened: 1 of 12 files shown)


app/models/content_item.rb
```diff
@@ -33,9 +34,6 @@ def self.create_or_replace(base_path, attributes)
         type: "conflict",
         code: "409",
         message: e.message,
-        fields: {
-          transmitted_at: [e.message],
```
kevindew: Should this have payload_version as a field now? I assume it's not a "need" since it's survived this long with transmitted_at which could be wrong but just wondering if it's a nice to have?

emmabeynon: Not too sure, but it doesn't seem to hurt, so adding in 👍

emmabeynon: Actually after a bit further investigation, it doesn't seem that `fields` is used anywhere in particular, and it would just contain the same error message as in the `message`, so not sure if there's any value in keeping it in?

PR #289, opened 2017-05-09
Reliably register routes

thomasleese:
Register routes more reliably by trying multiple times and rolling back the content item if it eventually fails.

[Trello Card](https://trello.com/c/1M8bgdQ0/928-ensure-the-content-store-reliably-updates-the-router-api-3)

### DIFF
--- a/app/models/content_item.rb
+++ b/app/models/content_item.rb
@@ -2,6 +2,11 @@ class ContentItem
   include Mongoid::Document
   include Mongoid::Timestamps::Updated
 
+  def self.revert(previous_item:, item:)
+    item.remove unless previous_item
+    previous_item&.upsert
+  end
+
   def self.create_or_replace(base_path, attributes)
     previous_item = ContentItem.where(base_path: base_path).first
     lock = UpdateLock.new(previous_item)
@@ -15,7 +20,14 @@ class ContentItem
     item.assign_attributes(attributes)
 
     if item.upsert
-      item.register_routes(previous_item: previous_item)
+      begin
+        item.register_routes(previous_item: previous_item)
+      rescue => e
+        revert(previous_item: previous_item, item: item)
+        raise unless e.is_a?(GdsApi::BaseError)
+        item.errors.add(:routes, "Could not communicated with router.")
+        result = false
+      end
     else
       result = false
     end
--- a/app/models/content_item.rb
+++ b/app/models/content_item.rb
@@ -152,8 +152,7 @@ class ContentItem
   end
 
   def register_routes(previous_item: nil)
-    return if self.schema_name.start_with?("placeholder")
-    return if previous_item && previous_item.route_set == self.route_set
+    return unless should_register_routes?(previous_item: previous_item)
     self.route_set.register!
   end
 
@@ -168,6 +167,12 @@ protected
   end
 
 private
+
+  def should_register_routes?(previous_item: nil)
+    return false if self.schema_name.start_with?("placeholder")
+    return false if previous_item && previous_item.route_set == self.route_set
+    true
+  end
 
   def authorised_user_uids
     access_limited.fetch('users', [])
--- a/app/models/content_item.rb
+++ b/app/models/content_item.rb
@@ -153,7 +153,14 @@ class ContentItem
 
   def register_routes(previous_item: nil)
     return unless should_register_routes?(previous_item: previous_item)
-    self.route_set.register!
+
+    tries = Rails.application.config.register_router_retries
+    begin
+      route_set.register!
+    rescue GdsApi::BaseError
+      tries -= 1
+      tries.positive? ? retry : raise
+    end
   end
 
   def base_path_without_root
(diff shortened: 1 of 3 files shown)


app/models/content_item.rb
```diff
@@ -152,9 +163,17 @@ def includes_auth_bypass_id?(auth_bypass_id)
   end
 
   def register_routes(previous_item: nil)
-    return if self.schema_name.start_with?("placeholder")
-    return if previous_item && previous_item.route_set == self.route_set
-    self.route_set.register!
+    return unless should_register_routes?(previous_item: previous_item)
+
+    3.times do
```
steventux: I'd be inclined to make this number a constant or put it in config if only to give it a label.

steventux: Is this retry / raise behaviour covered in the model spec?

app/models/content_item.rb
```diff
@@ -152,9 +163,17 @@ def includes_auth_bypass_id?(auth_bypass_id)
...
+      begin
+        route_set.register!
+        break
+      rescue GdsApi::BaseError => e
+        last_exception = e
+      end
+      raise last_exception
```
bevanloon: If `route_set.register!` raises an exception the first time, is the intention to retry 2 more times? In which case, do we want a test in the test suite to ensure that we are retrying?

The underlying question as a ruby neophyte - does `raise` break out of `3.times` prematurely?

thomasleese: I looked into how we could test this, unfortunately it's currently not possible to configure WebMock to return a different status code when calling again.

I took the code from https://github.com/jordansissel/software-patterns/tree/master/retry-on-failure/ruby, I assume that `raise` does not break out of the block if `last_exception` is `nil`.

app/models/content_item.rb
```diff
@@ -2,6 +2,11 @@ class ContentItem
   include Mongoid::Document
   include Mongoid::Timestamps::Updated
 
+  def self.revert(previous_item:, item:)
+    item.remove unless previous_item
+    previous_item&.upsert
```
steventux: I've never seen this before what's with the `&.upsert`?

app/models/content_item.rb
thomasleese: I think I'll refactor this code out so there is a method `retry` which does the retrying, and test that separately.

steventux: This raise could be conditional on a `times` block index var.

app/models/content_item.rb
thomasleese: The code used to be:

```
if previous_item
  previous_item.upsert
else
  item.remove
end
```

But Robocop complained that I should be using the `&.` (only calls the method if the operand is not `nil`) operating instead. Personally I think this code is less clear, but it was the only way to make Robocop happy.

thomasleese:
@bevanloon @steventux I believe I have addressed the issues you raised.

Regarding the `revert`, there isn't a chance of the content item being lost because it only calls `.remove` if there wasn't any existing item anyway, the only possibility is that we end up in the same inconsistent state, but now it happens a lot more rarely.

app/models/content_item.rb
```diff
@@ -15,7 +20,13 @@ def self.create_or_replace(base_path, attributes)
     item.assign_attributes(attributes)
 
     if item.upsert
-      item.register_routes(previous_item: previous_item)
+      begin
+        item.register_routes(previous_item: previous_item)
+      rescue GdsApi::BaseError
```
kevindew: I'd imagine we'd want to revert if any exception is thrown here? could be that we just re-raise it if it's not an instance of `GdsApi::BaseError`?

thomasleese: Yes, I agree!

app/models/content_item.rb
```diff
@@ -152,9 +164,16 @@ def includes_auth_bypass_id?(auth_bypass_id)
...
-    self.route_set.register!
+    return unless should_register_routes?(previous_item: previous_item)
+
+    tries = Rails.application.config.register_router_retries
+    begin
+      route_set.register!
+    rescue GdsApi::BaseError
+      tries -= 1
+      retry if tries.positive?
+      raise
```
kevindew: Sorry - super minor, but might be nice for this to be `tries.positive? ? retry : raise`

app/models/content_item.rb
```diff
@@ -15,7 +20,14 @@ def self.create_or_replace(base_path, attributes)
     item.assign_attributes(attributes)
 
     if item.upsert
-      item.register_routes(previous_item: previous_item)
+      begin
+        item.register_routes(previous_item: previous_item)
+      rescue => e
+        raise unless e.is_a?(GdsApi::BaseError)
+        revert(previous_item: previous_item, item: item)
```
kevindew: Didn't we want to revert before the raise?

kevindew:
This looks good to me, just the minor comments. Happy to approve though. 

@cbaines just wondering if you had any thoughts since you were working with this in the first place?

cbaines:
> @cbaines just wondering if you had any thoughts since you were working with this in the first place?

I haven't looked at the code too closely, but I think the idea is good :) Making the update transactional will help reduce inconsistencies between the router and content-store.

thomasleese:
@kevindew Thanks, I've addressed your comments. 😄

PR #294, opened 2017-05-23
Increase publish intent cache leeway

gpeng:
This commit increases the cache leeway for `PublishIntent` to 5 minutes. This has the effect of setting the max-age headers to 0 for content that has a publish intent for responses to requests received for up to 5 minutes after the scheduled publication time.

This change is being made following a perceived delay in the publishing of a statistics publication from Whitehall. Examination of the logs show that the publish request was initially sent from Whitehall within 15 seconds of the scheduled publication time but was held up, probably due
to a backlog of sidekiq jobs in the downstream queue at publishing api, and didn't arrive at content-store until over 2 minutes after the scheduled time. This caused users that tried to access the content during that time to be served the default 30 minute max-age and the publication of the content to appear to have been delayed for 30 minutes. 

This will be part of further mitigating changes that may include creating a higher priority queue for scheduled publishing. 

[Trello](https://trello.com/c/xfIvn3Pv/133-statistics-announced-in-advance-don-t-always-display-on-time)

### DIFF
--- a/app/models/publish_intent.rb
+++ b/app/models/publish_intent.rb
@@ -21,7 +21,7 @@ class PublishIntent
     ::FindByPath.new(self).find(path)
   end
 
-  PUBLISH_TIME_LEEWAY = 1.minute
+  PUBLISH_TIME_LEEWAY = 5.minutes
 
   field :_id, as: :base_path, type: String, overwrite: true
   field :publish_time, type: DateTime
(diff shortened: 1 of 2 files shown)


gpeng:
Thanks both... I did have a think about some way where the actual publishing cancelled the cache reduction but that was going to get a bit complicated to solve a support ticket. I think that would be a better solution. Maybe pass an `intent_id` with the publish request or something like that we could use to reset the cache time.

PR #296, opened 2017-06-14
[Do not merge] Downgrade Airbrake gem

leenagupte:
Trello card: https://trello.com/c/wg6F6Gk3

The inconsistent document checker task is no longer sending an error report to Errbit. 
This seems to have started happening when the app was upgraded to Rails 5. As it's known that version 5 of Airbrake is not compatible with Rails 5, the hope is that simply changing the version of Airbrake used to the alphagov fork will fix the problem.


thomasleese:
Closing since this doesn't solve the problem.

PR #297, opened 2017-06-16
Log the delay in scheduled publishing

gpeng:
We need data on the time it takes for scheduled publications to reach
the frontend to ensure that we are publishing sufficiently quickly to
meet statuatory requirements.

This PR records the difference between intended publish time and the
actual time that content arrives at content store to statsd.

In order to only log the latency for content submissions that relate to
a PublishIntent, we also delete the PublishIntent if we match one. This
prevents unscheduled submissions later in the day from also recording a
latency. This would give us incorrect data as the latency could be as
long as hours.

### DIFF
--- a/app/controllers/content_items_controller.rb
+++ b/app/controllers/content_items_controller.rb
@@ -27,6 +27,15 @@ class ContentItemsController < ApplicationController
       ContentItem.create_or_replace(encoded_base_path, @request_data)
     end
 
+    intent = PublishIntent.find_by_path(encoded_base_path)
+    if intent.present? && intent.publish_time.past?
+      intent.destroy
+      Rails.application.statsd.timing(
+        "scheduled_publishing_delay.#{item.document_type}",
+        ((item.updated_at.to_time - intent.publish_time.to_time) * 1000).to_i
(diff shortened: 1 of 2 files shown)


boffbowsh:
Due to the way aggregation in statsd works, edits made later in the day could massively inflate the reported values. De-duping the content ids or paths using Redis might be a good solution, however content-store doesn't currently talk to Redis. We could also just add it as an attribute on the `ContentItem` record.

Alternatively is there a reason we don't simply delete the `PublishIntent` once the content item has been published?

boffbowsh:
Added the Publish Intent deletion behaviour and a test. The test uncovered some issues in the time calculation (can't diff `DateTime`s, needs to be in integer milliseconds, subtraction wrong way round) so I fixed them too.

I haven't yet tested this locally as my development environment is out of date, so will test the branch on Integration on Monday.

boffbowsh:
Added. And tested on integration - it works fine.

boffbowsh:
They're deleted daily anyway, right? So if anything we should remove that cronjob a day after this is all deployed and working.

PR #299, opened 2017-06-23
Support base_path less content

gpeng:
Some link types are base_path less e.g. contact, government and world_location. Presenting `null` `api_path`, `web_url` and `api_url` for these formats makes them invalid with the frontend schemas.

This commit removes those fields when their value is null.

### DIFF
--- a/app/presenters/expanded_links_presenter.rb
+++ b/app/presenters/expanded_links_presenter.rb
@@ -14,7 +14,7 @@ class ExpandedLinksPresenter
           api_url: api_url(link),
           web_url: web_url(link),
           links: link[:links].present? ? self.class.new(link[:links]).present : {}
-        )
+        ).compact
       end
     end
   end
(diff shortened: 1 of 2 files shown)


kevindew:
I think we still end up with invalid schemas despite these changes. Since [base_path is a required field](https://github.com/alphagov/govuk-content-schemas/blob/master/dist/formats/news_article/frontend/schema.json#L378)

And with [additional properties being false](https://github.com/alphagov/govuk-content-schemas/blob/master/dist/formats/news_article/frontend/schema.json#L374) I guess any types that have additional fields end up being invalid.

Sorry...

gpeng:
😢 @kevindew I've made `base_path` optional now for this link type, well, I will do if I can ever find the magic combination to allow me to get CI to pass. https://github.com/alphagov/govuk-content-schemas/pull/629 Does that work?

kevindew:
@gpeng yup that looks good. What do you think about also making additionalProperties to true for both sets of link types?

gpeng:
Sure... I'll sort that out. Thanks for your help :+1:

PR #306, opened 2017-08-17
Add rake task to validate a sample of content

tijmenb:
The content-store currently returns a lot of invalid items. Before we start fixing that we need some way to track the number of invalid things.

This commit adds a rake task that is meant to be run periodically. It will take 100 items for each schema and validate those. The result is outputted so that we can see the errors in Jenkins. The overall result is sent to statsd as a gauge. This means we can generate a graph of the (sampled) percentage of content items that are invalid.

Ideally we would check every item in the content-store, but this is prohibitively expensive (it takes about 4 hours). The sample will hopefully give us enough data to start working on improving the state of the data.

https://trello.com/c/FesP8XSc

### DIFF
--- a/lib/tasks/validate_content_against_schema.rake
+++ b/lib/tasks/validate_content_against_schema.rake
@@ -1,0 +1,56 @@
+task validate_content_against_schema: :environment do
+  GovukContentSchemaTestHelpers.configure do |config|
+    config.schema_type = 'frontend'
+    config.project_root = Rails.root
+  end
+
+  schema_names = ContentItem.distinct(:schema_name).select(&:present?)
+  puts "Will check #{schema_names}"
+
+  reports = schema_names.map do |schema_name|
+    validate_content_for_schema(schema_name)
+  end
+
+  total_checked = reports.sum { |r| r[:counts][:checked] }
+  total_fail = reports.sum { |r| r[:counts][:fail] }
+  failure_percentage = ((total_fail / total_checked.to_f) * 100).to_i
+
+  Rails.application.statsd.gauge("document_validation.checked", total_checked)
+  Rails.application.statsd.gauge("document_validation.fail", total_fail)
+  Rails.application.statsd.gauge("document_validation.failure_percentage", failure_percentage)
+
+  puts reports.to_yaml
+end
+
+def validate_content_for_schema(schema_name)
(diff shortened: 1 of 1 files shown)


lib/tasks/validate_content_against_schema.rake
```diff
@@ -0,0 +1,56 @@
...
+  end
+
+  schema_names = ContentItem.distinct(:schema_name).select(&:present?)
+  puts "Will check #{schema_names}"
+
+  reports = schema_names.map do |schema_name|
+    validate_content_for_schema(schema_name)
+  end
+
+  total_checked = reports.reduce(0) { |s, r| s + r[:counts][:checked] }
```
thomasleese: Can we do `reports.sum { |r| r[:counts][:checked] }`?

lib/tasks/validate_content_against_schema.rake
```diff
@@ -0,0 +1,56 @@
...
+  # content_id ASC makes the sample semi-random
+  items = ContentItem.order('content_id ASC').where(schema_name: schema_name).limit(10)
+
+  counts = { checked: 0, okay: 0, fail: 0 }
+  errors = []
+
+  items.each do |content_item|
+    print '.'
+
+    counts[:checked] = counts[:checked] + 1
```
thomasleese: Should be possible to do `counts[:checked] += 1` and for the various times below.

lib/tasks/validate_content_against_schema.rake
```diff
@@ -0,0 +1,56 @@
...
+  Rails.application.statsd.gauge("document_validation.failure_percentage", failure_percentage)
+
+  puts reports.to_yaml
+end
+
+def validate_content_for_schema(schema_name)
+  api_url_callable = lambda { |base_path| "http://api.example.com/content#{base_path}" }
+
+  # content_id ASC makes the sample semi-random
+  items = ContentItem.order('content_id ASC').where(schema_name: schema_name).limit(10)
```
thomasleese: I would do `order(content_id: :asc)` here.

tijmenb:
@thomasleese cheers, updated!

tijmenb:
Resulting job on Jenkins:

https://deploy.integration.publishing.service.gov.uk/job/check-content-store

Integration dashboard:

https://grafana.integration.publishing.service.gov.uk/dashboard/db/content-validation-content-store

PR #309, opened 2017-08-23
Fix OpenAPI specification

thomasleese:
This fixes minor issues that were missed before.

[Trello Card](https://trello.com/c/WOxfVVNs/1021-2-add-existing-public-api-to-the-wip-documentation)

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -20,8 +20,12 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/ContentItem'
+        303:
+          description: The canonical path for this content item is a different path.
         404:
           description: No content item is available at that path.
+        410:
+          description: The content item is no longer available at that path.
 
 components:
   schemas:
@@ -48,6 +52,7 @@ components:
         - description
         - document_type
         - locale
+        - links
         - public_updated_at
         - schema_name
         - title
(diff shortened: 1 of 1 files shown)


openapi.yaml
```diff
@@ -65,6 +61,7 @@ components:
         content_id:
           type: string
           format: uuid
+          nullable: true
```
kevindew: Actually after some digging these can't be null (officially). It's hidden away obtusely [here](https://github.com/alphagov/publishing-api/blob/master/app/queries/get_edition_ids_with_fallbacks.rb#L11)

openapi.yaml
```diff
@@ -41,17 +45,9 @@ components:
...
-        - analytics_identifier
-        - api_path
-        - base_path
         - content_id
-        - description
-        - document_type
         - locale
-        - public_updated_at
-        - schema_name
         - title
```
kevindew: I've been trying to work out what is actually required here - and it's turned up to be a bit confusing (unsurprisingly). 

Looks like I was wrong about these being not required and instead may come out as nil:

```
[{:analytics_identifier=>nil,
  :api_path=>nil,
  :base_path=>nil,
  :content_id=>"c6e33b13-42a8-4431-980d-40bd55ec4518",
  :description=>nil,
  :document_type=>"contact",
  :locale=>"en",
  :public_updated_at=>"2016-05-23T14:22:13Z",
  :schema_name=>"contact",
  :title=>"Inward Investment Enquiries",
  :withdrawn=>false,
  :links=>{}}]
```

PR #312, opened 2017-09-01
Fall back to government-frontend instead of whitehall for gones

tuzz:
This was broken in Whitehall – no explanation or
alternative url was being shown. This is working
in government-frontend and we're trying to move
the frontend rendering logic there anyway.

This was discovered while doing QA for the
Whitehall rails upgrade:
https://trello.com/c/eg61wStW/4-upgrade-whitehall-to-a-supported-version-of-rails

### DIFF
--- a/app/models/content_item.rb
+++ b/app/models/content_item.rb
@@ -141,12 +141,12 @@ class ContentItem
     # For items that are registered in the content store which are not redirects
     # or gones we need to have a rendering_app. This is fine for all but the
     # "gone but not gone" exception defined in `gone?` where rendering_app is
-    # not part of the schema for a gone but it required to register the route.
+    # not part of the schema for a gone but is required to register the route.
     #
-    # This rather nastily fallsback to whitehall frontend for gones that are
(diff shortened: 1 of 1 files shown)


app/models/content_item.rb
```diff
@@ -141,12 +141,12 @@ def router_rendering_app
...
     # or gones we need to have a rendering_app. This is fine for all but the
     # "gone but not gone" exception defined in `gone?` where rendering_app is
-    # not part of the schema for a gone but it required to register the route.
+    # not part of the schema for a gone but is required to register the route.
     #
     # This rather nastily fallsback to whitehall frontend for gones that are
     # not gone and lack a rendering_app
     return rendering_app if schema_name != "gone" || gone?
-    rendering_app || "whitehall-frontend"
+    rendering_app || "government-frontend"
```
davidbasalla: Should we update the comment on line 146 that refers to falling back to whitehall as well?

tuzz: 👍  fixed

PR #328, opened 2017-11-07
Initial PaaS manifest

stephenharker:
The PaaS manifest file will not have any effect for now.

### DIFF
--- a/manifest.yml
+++ b/manifest.yml
@@ -1,0 +1,22 @@
+---
+applications:
+- name: srh-content-store
+  memory: 256M
+  buildpack: ruby_buildpack
+  instances: 1
+  services:
+  - content-store-mongo
+  env:
+    GOVUK_APP_NAME: content-store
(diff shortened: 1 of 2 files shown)


manifest.yml
```diff
@@ -0,0 +1,22 @@
...
+- name: govuk-dev-content-store
+  memory: 256M
+  buildpack: ruby_buildpack
+  instances: 1
+  services:
+  - content-store-mongo
+  env:
+    GOVUK_APP_NAME: content-store
+    GOVUK_CONTENT_SCHEMAS_PATH: /govuk-content-schemas
+    # I had to add these. I made up the values, no idea what they *should* be
```
MatMoore: These appear to be set here at the moment: https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk/manifests/deploy/config.pp

For the vagrant dev environment, app domain is "dev.gov.uk" and website root is  "http://www.dev.gov.uk".

It's probably not actually used for content store, but to be consistent with other apps it should use the environment specific domain.

PR #329, opened 2017-11-23
Handle deleting routes when deleting content items

cbaines:
Copied from the last commit:

This change provides a means to delete routes, from the Publishing
API, via the Content Store.

On an architectural level, the Publishing API doesn't directly talk to
the Router API, all communication goes through the Content Store. The
Publishing API can delete content items from the Content Store,
e.g. if you unpublish some content with a vanish type.

This change means that deletions from the Content Store will mean that
the Content Store will delete the relevant routes from the router.

One practical issue which this helps with is removing content/routes
when they are no longer wanted. This is tricky with content with
prefix routes. Say you have two bits of content, A and B where the
routes are:

  /A (type prefix)
  /A/B

If you wanted to remove B, such that A was responsible for /A/B, then
the vanish unpublishing type, with this change to the Content Store
would be a way of doing that.

This change doesn't address the issue of deleting routes when a
ContentItem no longer specifies that route, which I think is another
somewhat related deficiency.

### DIFF
--- a/app/models/route_set.rb
+++ b/app/models/route_set.rb
@@ -81,6 +81,21 @@ class RouteSet < OpenStruct
     commit_routes
   end
 
+  def delete!
+    return unless any_routes?
+
+    paths.each do |path|
+      begin
+        Rails.application.router_api.delete_route(path)
+      rescue GdsApi::HTTPNotFound
+        # Ignore this, as this path could have already been deleted
+        next
+      end
+    end
+
+    commit_routes
+  end
+
 private
 
   def register_rendering_app
@@ -114,6 +129,12 @@ private
 
   def commit_routes
     Rails.application.router_api.commit_routes
+  end
+
+  def paths
+    (routes + gone_routes + redirects).map do |hash|
+      hash[:path]
+    end
   end
 
   def any_routes?
--- a/app/models/route_set.rb
+++ b/app/models/route_set.rb
@@ -86,7 +86,7 @@ class RouteSet < OpenStruct
 
     paths.each do |path|
       begin
-        Rails.application.router_api.delete_route(path)
+        router_api.delete_route(path)
       rescue GdsApi::HTTPNotFound
         # Ignore this, as this path could have already been deleted
         next
@@ -99,11 +99,11 @@ class RouteSet < OpenStruct
 private
 
   def register_rendering_app
-    Rails.application.router_api.add_backend(rendering_app, Plek.find(rendering_app) + "/")
+    router_api.add_backend(rendering_app, Plek.find(rendering_app) + "/")
   end
 
   def register_redirect(route)
-    Rails.application.router_api.add_redirect_route(
+    router_api.add_redirect_route(
       route.fetch(:path),
       route.fetch(:type),
       route.fetch(:destination),
@@ -113,14 +113,14 @@ private
   end
 
   def register_gone_route(route)
-    Rails.application.router_api.add_gone_route(
+    router_api.add_gone_route(
       route.fetch(:path),
       route.fetch(:type),
     )
   end
 
   def register_route(route, rendering_app)
-    Rails.application.router_api.add_route(
+    router_api.add_route(
       route.fetch(:path),
       route.fetch(:type),
       rendering_app,
@@ -128,7 +128,11 @@ private
   end
 
   def commit_routes
-    Rails.application.router_api.commit_routes
+    router_api.commit_routes
+  end
+
+  def router_api
+    Rails.application.router_api
   end
 
   def paths
--- a/spec/integration/deleting_a_content_item_spec.rb
+++ b/spec/integration/deleting_a_content_item_spec.rb
@@ -6,12 +6,25 @@ RSpec.describe "Deleting a content item", type: :request do
   context "when the content item exists" do
     before do
       FactoryGirl.create(:content_item, base_path: base_path)
+
+      @delete_stubs = ContentItem.find_by(
+        base_path: base_path
+      ).routes.map do |route|
+        stub_route_deleted(route["path"])
+      end
     end
 
     it "deletes the content item" do
       delete "/content/vat-rates"
 
       expect(ContentItem.where(base_path: base_path).count).to eq(0)
+    end
+
+    it "deletes the routes" do
+      delete "/content/vat-rates"
+
+      @delete_stubs.each { |stub| assert_requested(stub, times: 1) }
+      assert_requested(stub_router_commit, times: 1)
     end
 
     it "returns a 200" do
--- a/app/controllers/content_items_controller.rb
+++ b/app/controllers/content_items_controller.rb
@@ -55,7 +55,11 @@ class ContentItemsController < ApplicationController
   end
 
   def destroy
-    ContentItem.find_by(base_path: encoded_base_path).destroy
+    content_item = ContentItem.find_by(base_path: encoded_base_path)
+
+    content_item.delete_routes
+    content_item.destroy
+
     render status: :ok
   end
 
--- a/app/models/content_item.rb
+++ b/app/models/content_item.rb
@@ -170,6 +170,12 @@ class ContentItem
     end
   end
 
+  def delete_routes
+    return unless should_register_routes?
+
+    route_set.delete!
+  end
+
   def base_path_without_root
     base_path&.sub(%r{^/}, "")
   end
--- a/spec/support/router_helpers.rb
+++ b/spec/support/router_helpers.rb
@@ -52,6 +52,13 @@ module RouterHelpers
       with(body: { "route" => hash_including("incoming_path" => path) })
     assert_not_requested(route_stub)
   end
+
+  def stub_route_deleted(path)
+    stub_http_request(
+      :delete,
+      "#{ROUTER_API_ENDPOINT}/routes?incoming_path=#{CGI.escape(path)}"
+    ).to_return(status: 200)
+  end
 end
 
 RSpec.configure do |config|


app/models/route_set.rb
```diff
@@ -81,6 +81,24 @@ def register!
...
+  def delete!
+    return unless any_routes?
+
+    paths =
+      routes.map { |route| route.fetch(:path) } +
+      redirects.map { |redirect| redirect.fetch(:path) }
+
+    paths.each do |path|
+      begin
+        Rails.application.router_api.delete_route(path)
```
thomasleese: Perhaps not related to this PR, but would be nice to get `Rails.application.router_api` into a method to avoid having to write `Rails.application` everywhere.

app/models/route_set.rb
```diff
@@ -81,6 +81,24 @@ def register!
     commit_routes
   end
 
+  def delete!
+    return unless any_routes?
+
+    paths =
```
thomasleese: Wonder if this should be extracted into a method.

spec/integration/deleting_a_content_item_spec.rb
```diff
@@ -6,6 +6,12 @@
   context "when the content item exists" do
     before do
       FactoryGirl.create(:content_item, base_path: base_path)
+
+      @delete_stubs = ContentItem.find_by(
+        base_path: base_path
+      ).routes.map do |route|
+        stub_route_deleted(route['path'])
```
thomasleese: Can we use double quotes here?

cbaines:
> Seems reasonable to me, would be nice to get f5627dc merged with the earlier commit.

Thanks for reviewing Tom, I've now squashed f5627dc in to the earlier commit.

app/models/route_set.rb
cbaines: Yeah, the separation between routes, gone_routes and redirects feels quite awkward to me. I've added a paths method to handle this now.

app/models/route_set.rb
cbaines: As you've commented about it, I've added separate commit making this change. I can always move it to a separate Pull Request if you'd like.

spec/integration/deleting_a_content_item_spec.rb
cbaines: Yep, I've changed it now.

kevindew:
🎉 this cleans an item off our backlog: https://trello.com/c/2vqZ6OFC/531-remove-routes-when-deleting-from-content-store - Thanks 👍

thomasleese:
It looks like this PR broke the end to end tests 😢 

https://ci.integration.publishing.service.gov.uk/job/publishing-e2e-tests/job/test-against/1279/console

```
  1) Discarding a draft on Specialist Publisher Discarding a draft CMA case
     Failure/Error: raise "Aborting reloading #{url} as a #{code} was returned"

     RuntimeError:
       Aborting reloading http://draft-origin.dev.gov.uk/cma-cases/tirra-lirra-by-the-river-1511452458?cachebust=1511452458 as a 410 was returned
     # ./spec/support/retry_helpers.rb:62:in `block in reload_url_until_status_code'
     # ./lib/retry_while_false.rb:7:in `block in call'
     # ./lib/retry_while_false.rb:6:in `loop'
     # ./lib/retry_while_false.rb:6:in `call'
     # ./spec/support/retry_helpers.rb:72:in `retry_while_false'
     # ./spec/support/retry_helpers.rb:59:in `reload_url_until_status_code'
     # ./spec/specialist_publisher/discarding_draft_spec.rb:32:in `then_i_get_a_404_on_draft_gov_uk'
     # ./spec/specialist_publisher/discarding_draft_spec.rb:9:in `block (2 levels) in <top (required)>'
```

https://github.com/alphagov/publishing-e2e-tests/blob/master/spec/specialist_publisher/discarding_draft_spec.rb#L32

As a side note, I notice that we were able to merge in the PR with only one of the 2 checks passing (the failing one being the end to end tests) which doesn't seem right.

cbaines:
> It looks like this PR broke the end to end tests 😢
> 
> https://ci.integration.publishing.service.gov.uk/job/publishing-e2e-tests/job/test-against/1279/console
> 
>   1) Discarding a draft on Specialist Publisher Discarding a draft CMA case
>      Failure/Error: raise "Aborting reloading #{url} as a #{code} was returned"
> 
>      RuntimeError:
>        Aborting reloading http://draft-origin.dev.gov.uk/cma-cases/tirra-lirra-by-the-river-1511452458?cachebust=1511452458 as a 410 was returned
>      # ./spec/support/retry_helpers.rb:62:in `block in reload_url_until_status_code'
>      # ./lib/retry_while_false.rb:7:in `block in call'
>      # ./lib/retry_while_false.rb:6:in `loop'
>      # ./lib/retry_while_false.rb:6:in `call'
>      # ./spec/support/retry_helpers.rb:72:in `retry_while_false'
>      # ./spec/support/retry_helpers.rb:59:in `reload_url_until_status_code'
>      # ./spec/specialist_publisher/discarding_draft_spec.rb:32:in `then_i_get_a_404_on_draft_gov_uk'
>      # ./spec/specialist_publisher/discarding_draft_spec.rb:9:in `block (2 levels) in <top (required)>'
> 
> https://github.com/alphagov/publishing-e2e-tests/blob/master/spec/specialist_publisher/discarding_draft_spec.rb#L32

I did look at that, but I didn't see how it could be related to the Content Store change, and disregarded it as a flaky test. Anyway, I'll investigate further now.

kevindew:
Yeah that's weird it could be merged. Guess it's this [hard/soft delete](https://github.com/alphagov/router-api/blob/master/app/controllers/routes_controller.rb#L32-L35) behaviour?

Yay though if e2e tests caught an issue before it spread

cbaines:
> Yeah that's weird it could be merged.

if the branch protection for master was setup to require that Jenkins job, it was probably undone on mass by the govuk-saas-config thing (e.g. [1]).

1: https://github.com/alphagov/govuk-saas-config/blob/master/github/lib/configure_repo.rb#L66

> Guess it's this hard/soft delete behaviour?

Yeah good spot. From looking at that, and the router-api, I think its actually the expectation in the publishing-e2e-tests that needs updating, so I'll put a PR in for that.

kevindew:
Yeah I think you're right about the configuration, e2e tests are now unticked as required. Hmm not sure best way to deal with that - I wonder if that script can set all contexts rather than a specific one. Can see on content-schemas we only require one branch too: https://github.com/alphagov/govuk-content-schemas/settings/branches/master /cc @tijmenb 

> Yeah good spot. From looking at that, and the router-api, I think its actually the expectation in the publishing-e2e-tests that needs updating, so I'll put a PR in for that.

I'd wonder if it's that these deletes should be hard deletes? I'd have thought we want 404 to be returned on a route that has been deleted. I think that behaviour exists currently as the rendering apps are passing the 404 they get from content store back. So I'd worry that this actually would change behaviour and create 410s where 404s are expected.

cbaines:
> I'd wonder if it's that these deletes should be hard deletes? I'd have thought we want 404 to be returned on a route that has been deleted. I think that behaviour exists currently as the rendering apps are passing the 404 they get from content store back. So I'd worry that this actually would change behaviour and create 410s where 404s are expected.

The change in the router-api seems pretty explicit on this [1]. It does concern me though, mostly as it means that the state of the router-api database is not a derivation of the state of the respective Content Store, but instead builds up these implicit gone routes when routes are deleted...

I wonder if this is a hangover from a world when there was no such thing as unpublishing content through the Publishing API. I tempted to look at reverting [1]...

1: https://github.com/alphagov/router-api/commit/fbaacf95469f28ff13ef3f02df6a7d1ffa5da57f

cbaines:
> Yeah good spot. From looking at that, and the router-api, I think its actually the expectation in the publishing-e2e-tests that needs updating, so I'll put a PR in for that.

This is up now; https://github.com/alphagov/publishing-e2e-tests/pull/72

kevindew:
@cbaines I feel uneasy about changing the e2e tests for this as I'm worried that we might be just kicking the can of an issue further here. I'm a bit worried that now if someone unpublishes with type vanish they're going to get a 410?

cbaines:
> @cbaines I feel uneasy about changing the e2e tests for this as I'm worried that we might be just kicking the can of an issue further here. I'm a bit worried that now if someone unpublishes with type vanish they're going to get a 410?

Ok, yeah, this will mean that the end-to-end behaviour will change in the vanish case, something which I didn't anticipate as I had no idea that router-api did this soft delete thing. I'll try to think forward on this, as your right, just taking the immediate next logical steps might not be sufficient.

cbaines:
> Ok, yeah, this will mean that the end-to-end behaviour will change in the vanish case, something which I didn't anticipate as I had no idea that router-api did this soft delete thing. I'll try to think forward on this, as your right, just taking the immediate next logical steps might not be sufficient.

Right, getting rid of this soft deletion functionality in the Router API service might be good in the future, but its too complicated to do right now, the data needs looking at in more detail if we want to keep the behaviour of the site the same. 

So, maybe we can just use the hard_delete parameter to ignore this behaviour by default. We won't get any gone routes automatically when deleting routes, which could break things, but that scenario is more undefined and less likely than the definite change in the vanish unpublishing.

First thing is to add support to the gds-api-adapters https://github.com/alphagov/gds-api-adapters/pull/770

cbaines:
> So, maybe we can just use the hard_delete parameter to ignore this behaviour by default. We won't get any gone routes automatically when deleting routes, which could break things, but that scenario is more undefined and less likely than the definite change in the vanish unpublishing.

Right, hopefully the final follow up for this, as I think I'm close to digging out of this hole. I've finally managed to get a change prepared for passing the `hard_delete=true` query parameter to the Router API [1].

1: https://github.com/alphagov/content-store/pull/331

PR #330, opened 2017-11-25
Update Ruby to 2.4.2 and bump Gemfile dependancies

RoryMacDonald:
Ahead of enabling Dependabot on the repository, we want to be using the latest
dependancies, to minimise the onslaught of incremental PRs.


Gemfile
```diff
@@ -31,7 +31,7 @@ group :development, :test do
   gem 'rspec-rails', '~> 3.0'
   gem 'database_cleaner', '~> 1.6.1'
   gem 'factory_girl', '~> 4.4'
```
kevindew: No biggie, but might want to use this as a time to switch to [factory_bot](https://github.com/thoughtbot/factory_bot_rails) - could well be a separate PR though if it needs a bunch of renaming vars

cbaines:
> The issue with the end to end tests is outlined alphagov/publishing-e2e-tests#72 so we should for it to go green before merging.

On this, rebasing these patches on master should fix this now.

Gemfile
tijmenb: Probably best done in a separate PR?

RoryMacDonald:
Nice one, thanks @cbaines. Those fixes into master seem to have done the trick.

PR #331, opened 2017-11-27
Hard delete routes

cbaines:
Copied from the third commit:

This avoids the Router API service converting these routes to gone
routes, and gets it to actually delete them.

I think this behaviour is better, it avoids state building up in the
Router API and also means that the unpublishing type of vanish will
work, as the routes will be deleted from the router.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -19,7 +19,7 @@ gem 'hashdiff', require: false
 if ENV['GDS_API_ADAPTERS_DEV']
   gem 'gds-api-adapters', path: '../gds-api-adapters'
 else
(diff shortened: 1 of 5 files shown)


cbaines:
Build seems to be failing as it can't fetch version 50.0.0 of the gds-api-adapters... it worked on my machine :(

cbaines:
Installing the gems is now working, but the Publishing E2E tests are now failing, as the content on the 404 page is slightly different... I've got a Pull Request to remove this expectation here [1].

1: https://github.com/alphagov/publishing-e2e-tests/pull/73

PR #385, opened 2018-02-22
Record publishing delays in graphite

suzannehamilton:
Log the difference between the scheduled publishing time and the actual publishing time. Use a Statsd `timing` stat, which automatically calculates other stats like the mean and standard deviation.

It might be more useful to calculate some daily aggregates, but at least this will let us start recording some data about publishing delays.


suzannehamilton:
Failing tests. :( I'll fix.

PR #388, opened 2018-02-27
Report scheduled publishing delays to graphite

suzannehamilton:
Add a cron job which runs a rake task every night to send publishing delay stats to graphite.

Calculate the stats directly rather than using a stats gem because the available gems weren't very suitable:

- descriptive-statistics (https://github.com/jtescher/descriptive-statistics) doesn't calculate percentiles in the expected way: it returns nil rather than the maximum value if the percentile requested is larger than the data set, e.g. the 95th percentile of a data set with less than 20 values.
- SciRuby statsample (https://github.com/sciruby/statsample) is too large for what we need: it adds dependencies like a PDF writer and a visualisation toolkit.

### DIFF
--- a/lib/tasks/data_hygiene/publishing_delay_reporter.rb
+++ b/lib/tasks/data_hygiene/publishing_delay_reporter.rb
@@ -1,0 +1,18 @@
+class Tasks::DataHygiene::PublishingDelayReporter
+  def report
+    now = Time.zone.now
+
+    log_entries = ScheduledPublishingLogEntry.where(created_at: (now - 1.day)..now)
+    delays = log_entries.map(&:delay_in_milliseconds)
+
+    unless delays.empty?
+      GovukStatsd.gauge("scheduled_publishing.aggregate.mean_ms", mean(delays))
+    end
+  end
+
+private
+
+  def mean(delays)
+    delays.sum / delays.size.to_f
+  end
+end
--- a/lib/tasks/data_hygiene/publishing_delay_reporter.rb
+++ b/lib/tasks/data_hygiene/publishing_delay_reporter.rb
@@ -6,13 +6,24 @@ class Tasks::DataHygiene::PublishingDelayReporter
     delays = log_entries.map(&:delay_in_milliseconds)
 
     unless delays.empty?
-      GovukStatsd.gauge("scheduled_publishing.aggregate.mean_ms", mean(delays))
+      GovukStatsd.gauge("scheduled_publishing.aggregate.mean_ms", Stats.mean(delays))
+      GovukStatsd.gauge("scheduled_publishing.aggregate.95_percentile_ms", Stats.percentile(delays, 95))
     end
   end
+end
 
-private
+class Tasks::DataHygiene::PublishingDelayReporter::Stats
+  def self.mean(values)
+    raise ArgumentError.new("Cannot calculate the mean of an empty array") if values.empty?
 
-  def mean(delays)
-    delays.sum / delays.size.to_f
+    values.sum / values.size.to_f
+  end
+
+  def self.percentile(values, percentile)
+    raise ArgumentError.new("Cannot calculate percentile #{percentile} of an empty array") if values.empty?
+
+    ordinal = (values.length * percentile.to_f / 100).ceil
+    index = [0, ordinal - 1].max
+    values.sort[index]
   end
 end
--- a/config/schedule.rb
+++ b/config/schedule.rb
@@ -9,3 +9,7 @@ job_type :rake, "cd :path && govuk_setenv content-store bundle exec rake :task :
 every 1.day, at: "2:00 am" do
   rake "housekeeping:cleanup_publish_intents"
 end
+
+every 1.day, at: "2:15 am" do
+  rake "publishing_delay_report:report_delays"
+end
--- a/lib/tasks/data_hygiene/publishing_delay_report.rake
+++ b/lib/tasks/data_hygiene/publishing_delay_report.rake
@@ -1,0 +1,8 @@
+require "tasks/data_hygiene/publishing_delay_reporter"
+
+namespace :publishing_delay_report do
+  desc "Report on the delay between scheduled and actual publication times"
+  task report_delays: :environment do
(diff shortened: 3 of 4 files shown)


lib/tasks/data_hygiene/publishing_delay_reporter.rb
```diff
@@ -0,0 +1,32 @@
+class Tasks::DataHygiene::PublishingDelayReporter
+  def report
+    now = Time.now
```
h-lame: We should use `Time.zone.now` to make sure we take timezones into account (and I think Rails will do utc-ification on them when passing those into db queries).

lib/tasks/data_hygiene/publishing_delay_reporter.rb
```diff
@@ -0,0 +1,32 @@
...
+
+    log_entries = ScheduledPublishingLogEntry.where(
+      :scheduled_publication_time.gt => now - 1.day,
+      :scheduled_publication_time.lte => now
```
h-lame: Mongoid will _probably_ allow us to write `.where(scheduled_publication_time: (now - 1.day)..now)`.  It _might_ be more efficient?  I am not a mongo expert by any means.

lib/tasks/data_hygiene/publishing_delay_reporter.rb
```diff
@@ -0,0 +1,32 @@
...
+    )
+    delays = log_entries.map(&:delay_in_milliseconds)
+
+    unless delays.empty?
+      GovukStatsd.gauge("scheduled_publishing.aggregate.mean_ms", Stats.mean(delays))
+      GovukStatsd.gauge("scheduled_publishing.aggregate.95_percentile_ms", Stats.percentile(delays, 95))
```
h-lame: If we used GovukStatsd.timing would on all the delays would we get all these stats for free?  https://github.com/etsy/statsd/blob/master/docs/metric_types.md#timing  It might be too much though?

lib/tasks/data_hygiene/publishing_delay_reporter.rb
suzannehamilton: Good shout. As far as I can tell it runs exactly the same query in Mongo, but it's sub-millisecond in either case.

lib/tasks/data_hygiene/publishing_delay_reporter.rb
suzannehamilton: I might well be missing something (timezones are confusing), but won't you get the same result whether you use `Time.now` and `Time.zone.now` precisely because of Rails utc-ification?

If I understand it correctly, `Time.now` will get the current time in the system's timezone, and `Time.zone.now` will get the current time in Rails's timezone, but they both refer to the same instant in time. So it shouldn't matter which one you put into the DB query?

I've played about with timezones in a Rails console and couldn't find a query which gave different results.

lib/tasks/data_hygiene/publishing_delay_reporter.rb
h-lame: Oh, good to know it generates the same query.  Not that activerecord provides less-than or greater-than operators for where clauses, but I'm p. sure it wouldn't optimize two conditionals like this into a between query for you if it did.

lib/tasks/data_hygiene/publishing_delay_reporter.rb
h-lame: Maybe if our app and servers are all set to utc then it doesn't actually matter because the timezones are always UTC so no conversion actually happens.  It might be more of an issue in summertime when UK local time isn't UTC.  I agree it'd be minor though, as I don't think we'd miss out on any data.

I might be cargo-culting this without much internal debate just because I've had it drilled into me by loads of blog posts and presentations that in Rails we should always use `Time.zone` and never `Time`.  Loathe as I am to point to rubocop as an arbiter of "good", it does have a cop for that says the same: https://github.com/bbatsov/rubocop/blob/master/lib/rubocop/cop/rails/time_zone.rb, which we'd eventually fall foul of when/if we upgrade the version of govuk-lint/rubocop on this project.

lib/tasks/data_hygiene/publishing_delay_reporter.rb
suzannehamilton: Not exactly, because statsd only calculates the the stats within each reporting time interval. So you'll get data points for the mean and 90th percentile for the two or three docs that were published at the same time - it won't give you the average across an arbitrary interval like 24 hours.

But we are already recording the timing data (see #385) and it might be possible to use Graphite to calculate something useful from it. The stats won't be totally accurate because statsd has already done some aggregation, but they might be good enough. I've had a go at building a dashboard, and I'll see if Paul thinks it'll do the job: https://grafana.publishing.service.gov.uk/dashboard/db/publishing-delays-work-in-progress

lib/tasks/data_hygiene/publishing_delay_reporter.rb
suzannehamilton: OK, I've updated it. 👍 I don't think it'll affect this query, but I agree that it's better to use in general (certainly for anything user-facing).

lib/tasks/data_hygiene/publishing_delay_reporter.rb
suzannehamilton: Just spoken to Paul, and he'd like us to add this cron job so we have more accurate stats.

h-lame: Yeah, I wasn't sure if `timing` would be useful, but I'd only just discovered it and wanted to use the new shiny thing everywhere ;).  If it's not fit for purpose then let's not use it.

suzannehamilton:
@h-lame: I've just updated this PR with a slight change: it now aggregates stats for docs _published_ in the last 24 hours, rather than those _scheduled_ for the last 24 hours.

This means we don't lose the stats for any that were scheduled for around the same time as the cron job. Most scheduled publishing happens during working hours, but it does occasionally happen overnight.

h-lame:
@suzannehamilton - ah, yes, that makes sense. :+1:

PR #389, opened 2018-02-27
Add new content purpose supergroup and subgroup

vanitabarrett:
We have added `content_purpose_supergroup` and `content_purpose_subgroup` to govuk-document-types https://github.com/alphagov/govuk_document_types/pull/36. We need to update content-store with this change.

**Note: The tests will fail until https://github.com/alphagov/govuk-content-schemas/pull/722 is deployed.**

### DIFF
--- a/app/models/content_item.rb
+++ b/app/models/content_item.rb
@@ -61,6 +61,8 @@ class ContentItem
   field :format, type: String
   field :document_type, type: String
   field :content_purpose_document_supertype, type: String, default: ''
+  field :content_purpose_subgroup, type: String, default: ''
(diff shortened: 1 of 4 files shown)


app/models/content_item.rb
```diff
@@ -66,6 +66,8 @@ def self.find_by_path(path)
   field :navigation_document_supertype, type: String, default: ''
   field :search_user_need_document_supertype, type: String, default: ''
   field :user_journey_document_supertype, type: String, default: ''
+  field :content_purpose_supergroup, type: String, default: ''
```
leenagupte: Minor thing, but I'd put these in alphabetical order.

PR #394, opened 2018-03-01
Revert "Add scheduled publication time to content item response"

suzannehamilton:
This reverts commit cf1747115f9f00e3586c067fe014f6207ca91da5 (PR #391).

This increased the response times by around 50% on staging, so revert this while we consider how to improve this, e.g. by adding an index to ScheduledPublishingLogEntry.

Response times are 150-200 ms on integration, so the extra query didn't make requests noticeably slower.  But on staging, where response times are around 10 ms, the extra query made a significant difference (the red line marks the deployment to staging):

screen shot 2018-03-01 at 14 55 03

Responses are still fast, but since this is the content-store I'd like to at least investigate adding an index to avoid a 50% increase in response times.

### DIFF
--- a/app/controllers/content_items_controller.rb
+++ b/app/controllers/content_items_controller.rb
@@ -10,20 +10,13 @@ class ContentItemsController < ApplicationController
       PublishIntent.find_by_path(encoded_request_path)
     end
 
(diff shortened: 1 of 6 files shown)


boffbowsh:
👍 yay for monitoring performance at deploy time!

PR #396, opened 2018-03-02
Add scheduled publishing delay to the content item response

suzannehamilton:
This allows publishers to check the delay without having to compare it to the updated time, which may be misleading if the document has been manually published after the scheduled publishing.

The tests will fail until the new field is added to the content schemas (alphagov/govuk-content-schemas#726).

### DIFF
--- a/app/presenters/content_item_presenter.rb
+++ b/app/presenters/content_item_presenter.rb
@@ -46,7 +46,10 @@ class ContentItemPresenter
       "details" => RESOLVER.resolve(item.details),
     ).tap do |i|
       i["redirects"] = item["redirects"] if i["schema_name"] == "redirect"
-      i["publishing_scheduled_at"] = scheduled_publishing.scheduled_publication_time if scheduled_publishing
+      if scheduled_publishing
+        i["publishing_scheduled_at"] = scheduled_publishing.scheduled_publication_time
+        i["scheduled_publishing_delay_seconds"] = scheduled_publishing.delay_in_milliseconds / 1000
+      end
     end
   end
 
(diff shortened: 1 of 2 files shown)


app/presenters/content_item_presenter.rb
```diff
@@ -46,7 +46,10 @@ def as_json(options = nil)
       "details" => RESOLVER.resolve(item.details),
     ).tap do |i|
       i["redirects"] = item["redirects"] if i["schema_name"] == "redirect"
-      i["publishing_scheduled_at"] = scheduled_publishing.scheduled_publication_time if scheduled_publishing
+      if scheduled_publishing
+        i["publishing_scheduled_at"] = scheduled_publishing.scheduled_publication_time
+        i["scheduled_publishing_delay_seconds"] = scheduled_publishing.delay_in_milliseconds / 1000
```
cbaines: It might be neater to add this in to the content item at the point it's added to the Content Store. As far as I can tell, this won't change after it's been computed at the time the content item is added/updated, and this would keep the logic around the scheduled publishing behaviour together more in the `ContentItemsController` where I can see it's already involved. What do you think?

suzannehamilton: I've had a go in #400. It's a bit awkward because we need to make sure the publishing delays are preserved when a new version of the document is published, but it's probably worth it to avoid the extra query on each content store request. What do you think?

cbaines: We just talked about this, but following up on GitHub for completeness. I'm ok with this or the alternative PR you proposed.

suzannehamilton:
After talking to @cbaines, I'm going to close this PR in favour of #400.

PR #400, opened 2018-03-08
Make publishing time a DB field and add publishing delay

suzannehamilton:
This is an alternative implementation of #396.

- As suggested in #396, move the `publishing_scheduled_at` field to a database field rather than adding it at query time. This saves the content-store from searching the scheduled publishing log entries on every GET request.
- Add a new `scheduled_publishing_delay_seconds` field to allow publishers to check the delay without having to compare it to the updated time, which may be misleading if the document has been manually published after the scheduled publishing.

cc @cbaines

### DIFF
--- a/app/controllers/content_items_controller.rb
+++ b/app/controllers/content_items_controller.rb
@@ -140,14 +140,12 @@ private
       .first
 
     if new_scheduled_publishing?(intent, document_type, latest_log_entry)
-      log_entry = ScheduledPublishingLogEntry.create(
+      latest_log_entry = ScheduledPublishingLogEntry.create(
         base_path: base_path,
         document_type: document_type,
         scheduled_publication_time: intent.publish_time,
       )
-      GovukStatsd.timing("scheduled_publishing.delay_ms", log_entry.delay_in_milliseconds)
-
-      log_entry
+      GovukStatsd.timing("scheduled_publishing.delay_ms", latest_log_entry.delay_in_milliseconds)
     end
 
     latest_log_entry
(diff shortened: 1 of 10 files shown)


suzannehamilton:
@cbaines I've updated the PR as you suggested - it now fetches the latest log entry rather than copying data between versions of content items.

app/controllers/content_items_controller.rb
```diff
@@ -10,32 +10,26 @@ def show
...
-      )
+      render json: ContentItemPresenter.new(item, api_url_method), status: http_status(item)
     else
       render json_forbidden_response
     end
   end
 
   def update
+    intent = PublishIntent.find_by_path(encoded_base_path)
+    log_entry = find_or_create_scheduled_publishing_log(encoded_base_path, @request_data["document_type"], intent)
```
cbaines: As far as I can see, for most content, which I assume isn't published with a prior intent, this will be `nil`? If that is the case, it might be good to incorporate that in to the name somehow?

suzannehamilton: I'm struggling to come up with a non-clunky name. The best options so far are `optional_scheduled_publishing_log` or `optional_publishing_log`, but I'm worried it implies that you can opt into creating the log, when really it gets created for a specific type of document.

Any better suggestions?

cbaines: The only suggestion I have would be to make the name more obscure, e.g. something like `scheduled_publishing_log_entry`, which at least prevents any assumptions being made about what the method does from the name.

I'm quite happy to go with the existing name though.

PR #401, opened 2018-03-08
Group scheduled publishing delay stats by document type

suzannehamilton:
Report statistics and statistics announcements separately to other publication types:

- It's important for national and official statistics to be published on time, so monitor these separately
- Statistics announcements can have incorrect publishing delays. In most cases they are reported correctly, but sometimes they can appear to be published hours or days late because a publisher has linked an announcement to a statistics set _after_ the statistics are published.

### DIFF
--- a/lib/tasks/data_hygiene/publishing_delay_reporter.rb
+++ b/lib/tasks/data_hygiene/publishing_delay_reporter.rb
@@ -1,13 +1,39 @@
 class Tasks::DataHygiene::PublishingDelayReporter
+  STATISTICS_TYPES = %w(
+    national_statistics
+    official_statistics
+    national_statistics_announcement
+    official_statistics_announcement
+  ).freeze
+
   def report
     now = Time.zone.now
 
     log_entries = ScheduledPublishingLogEntry.where(created_at: (now - 1.day)..now)
-    delays = log_entries.map(&:delay_in_milliseconds)
+    stats_reporter = StatsReporter.new(log_entries)
 
-    unless delays.empty?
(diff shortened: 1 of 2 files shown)


lib/tasks/data_hygiene/publishing_delay_reporter.rb
```diff
@@ -1,13 +1,39 @@
...
+  end
+
+  class StatsReporter
+    attr_reader :log_entries
+
+    def initialize(log_entries)
+      @log_entries = log_entries
+    end
+
+    def report(namespace)
```
h-lame: This is extremely minor, but could we avoid the slightly awkward `{ true }` argument when we want to report all the docs?  Something like:

```ruby
def report(namespace, &filter)
  delays = log_entries.select(&filter).map(&:delay_in_milliseconds)
  # rest of method...
end
```

Seems that if you pass a nil block to select it becomes a no-op.  I found this out in irb as follows:

```ruby
def foo(&block)
  [1,2,3].select(&block).map { |x| puts x }
end

irb(main):005:0> foo { true }
1
2
3
=> [nil, nil, nil]
irb(main):006:0> foo
1
2
3
=> [nil, nil, nil]
```

I thought we'd need to do a conditional, but it seems `select` is happy with a nil block and acts like `each` because every element is selected.  I'm like 50:50 on this being clear though, as I found it a little surprising, so entirely up to you.

suzannehamilton: I think that is neater without the `{ true }`. 👍 

Hopefully the purpose of `stats_reporter.report("all_document_types")` is clear enough, and it is unit tested.

PR #426, opened 2018-04-11
Delete flaky assertion

tijmenb:
This test failed on CI:

https://ci.integration.publishing.service.gov.uk/job/content-store/job/remove-supertypes/1/console

Flaky tests should be removed to avoid losing trust in test suites. I'll put something on someones board for a permanent fix.


cbaines:
> Flaky tests should be removed to avoid losing trust in test suites. I'll put something on someones board for a permanent fix.

Removing this does worry me a little, as I can't actually see from the test what might be causing it to be flaky...

PR #430, opened 2018-04-27
Remove all the code related to need_ids

cbaines:
The use of this field has been replaced with the `meets_user_needs`
link type.


cbaines:
The first step, is to make them optional in the `govuk-content-schemas`, which this PR does https://github.com/alphagov/govuk-content-schemas/pull/754

PR #434, opened 2018-05-02
Avoid false positives with publishing-api delay seconds

Davidslv:
I want to avoid more scenarios like the following on CI:

```
  1) content item write API with an earlier scheduled publishing log
entry adds the latency to the content item
     Failure/Error:
expect(content_item.scheduled_publishing_delay_seconds).to eq(25)

       expected: 25
            got: 24

       (compared using ==)
     # ./spec/integration/submitting_content_item_spec.rb:186:in `block
     # (3 levels) in <top (required)>''`

```

### DIFF
--- a/spec/integration/submitting_content_item_spec.rb
+++ b/spec/integration/submitting_content_item_spec.rb
@@ -183,7 +183,7 @@ describe "content item write API", type: :request do
 
       content_item = ContentItem.find_by(base_path: "/vat-rates")
       expect(content_item.publishing_scheduled_at).to be_within(0.001.seconds).of(publish_time)
-      expect(content_item.scheduled_publishing_delay_seconds).to eq(25)
+      expect(content_item.scheduled_publishing_delay_seconds).to be_between(24, 25).inclusive
     end
   end
 


spec/integration/submitting_content_item_spec.rb
```diff
@@ -183,7 +183,7 @@
 
       content_item = ContentItem.find_by(base_path: "/vat-rates")
       expect(content_item.publishing_scheduled_at).to be_within(0.001.seconds).of(publish_time)
-      expect(content_item.scheduled_publishing_delay_seconds).to eq(25)
+      expect(content_item.scheduled_publishing_delay_seconds).to be_between(0, 25).inclusive
```
thomasleese: It would be better to do `be_between(24, 25).inclusive` because if it's 0 then something is broken.

cbaines:
I'd prefer removing this (e.g. https://github.com/alphagov/content-store/pull/426 ) rather than weakening it. The test explicitly sets, and at least currently expects this value, so it seems very odd to allow it to differ.

Davidslv:
@cbaines I don't want to weaken the test, I think that @thomasleese suggestion makes more sense than my one, would you agree? I've changed the test to `be_between(24, 25).inclusive`

PR #438, opened 2018-05-08
Add a scheduled publication delay report

thomasleese:
This report shows content items with a delay in the scheduled publication for the last week. It outputs its results as a CSV file to standard output. The plan is for this to be automatically sent to various people every week.

[Trello Card](https://trello.com/c/rkOiRFUV/128-run-the-statistics-publication-delay-report-weekly-and-send-it-to-alice-2)

### DIFF
--- a/lib/publication_delay_report.rb
+++ b/lib/publication_delay_report.rb
@@ -1,0 +1,48 @@
+class PublicationDelayReport
+  def initialize(file)
+    @file = file
+  end
+
+  def call
+    write_csv
+  end
+
+  def self.call(*args)
+    new(*args).call
+  end
+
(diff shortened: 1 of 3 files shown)


lib/publication_delay_report.rb
```diff
@@ -0,0 +1,42 @@
+class PublicationDelayReport
+  def call
+    write_csv($stdout)
```
boffbowsh: I think it's probably better in this case to pass a target into `#call`, and have the Rake task pass `$stdout` in. Tiny nitpick ;)

thomasleese: Sure! I've amended and pushed that

PR #499, opened 2018-12-12
Set Mongo client to prefer reads from secondaries

stephenharker:
This will spread the read requests across the secondary Mongo instances
automatically and spare the primary from taking all the load as is
happening currently. This should obviate the need to increase the sizes
of the Mongo instances and should also improve performance.


stephenharker:
I suspect this was switched to "prefer_primary" when the problem of a third of requests going to the DR replica in the DR environment with extra high-latency penalty was discovered. We don't have this in AWS so it should be fine to read from any secondaries there.

PR #535, opened 2019-02-22
Create generic version of content store search

keithiopia:
Repurpose Deborah and Ruben's code that searches for gsi domain email addresses and dumps the results in a csv. Should be easy to plug in a specific search term, in this first case "another EU country".


lib/find_specific_terms.rb
```diff
@@ -0,0 +1,67 @@
...
+
+  private_class_method :new
+
+private
+
+  def write_csv
+    CSV.open("#{Rails.root}/tmp/search_term_content_items.csv", 'wb') do |csv|
+      csv << CSV_HEADERS
+
+      searchterm = %(another EU country)
```
rubenarakelyan: If you want a single string search term, you could use single or double quotes instead. This also means you don't need the `searchterm.each do |term|` block below - the contents of the block can just use `searchterm` directly.

rubenarakelyan: Also, `searchterm` could be provided each time this class is instantiated so that you can provide your own search term when you call this class via a rake task rather than having to change it in the code.

lib/tasks/search-term-report.rake
```diff
@@ -0,0 +1,11 @@
+namespace :report do
+  desc "Run a publication delay report of the last week"
+  task publication_delay_report: :environment do
+    PublicationDelayReport.call($stdout)
+  end
+
+  desc "Find all references to search term"
+  task task :find_specific_term_references, %i[term] => [:environment] do |_, args|
```
thomasleese: There are two `task`s here.

lib/tasks/search-term-report.rake
```diff
@@ -0,0 +1,11 @@
+namespace :report do
+  desc "Run a publication delay report of the last week"
+  task publication_delay_report: :environment do
+    PublicationDelayReport.call($stdout)
+  end
+
+  desc "Find all references to search term"
```
thomasleese: We can move this task into the `report.rake` file.

keithiopia: Could do, but we only really want to run one or the other (we don't want the gsi report every time)

lib/tasks/search-term-report.rake
keithiopia: Dagnabbit!

lib/tasks/search-term-report.rake
thomasleese: What you can do is add

```
desc "Find all references to search term"
task :find_specific_term_references, %i[term] => [:environment] do |_, args|
  FindSpecificTerm.call(args[:term])
end
```

into the file `report.rake` under the `namespace :report` and that allows people to run only the task `report:find_specific_term_references` without having to run the GSI task as well. And it means we can get rid of this file.

keithiopia: Done and done.

PR #593, opened 2019-07-17
Support access limiting by user organisation

theseanything:
This PR adds support for allowing users to see content restricted to organisations.

It compares the `X-Govuk-Authenticated-User-Organisation` header sent in the request to the organisations listed within the access limiting properties.

### DIFF
--- a/app/controllers/content_items_controller.rb
+++ b/app/controllers/content_items_controller.rb
@@ -16,7 +16,7 @@ class ContentItemsController < ApplicationController
     return error_404 unless item
     return redirect_canonical(item) if item.base_path != encoded_request_path
 
-    if can_view(item)
+    if can_view?(item)
       render json: ContentItemPresenter.new(item, api_url_method), status: http_status(item)
     else
       render json_forbidden_response
@@ -64,44 +64,25 @@ private
     redirect_to route, status: 303
   end
 
-#  def can_view(item)
-#    auth_user_id != 'invalid' && (check_bypass_id(item) || check_user_id(item) || not_access_limited(item))
-#  end
+  def can_view?(item)
+    return item.valid_bypass_id?(auth_bypass_id) if auth_bypass_id
 
-  def can_view(item)
-    if valid_user?(auth_user_id)
-        if item.access_limited?
-          item.viewable_by_bypass_id?(auth_bypass_id) || item.viewable_by_user_id?(auth_user_id)
-        else
-          true
-        end
-    else
-      false
-    end
-  end
-
-  def valid_user?(user_id)
-    user_id != 'invalid'
+    item.user_access?(
+      user_id: auth_user_id,
+      user_organisation_id: auth_organisation_id
+    )
   end
 
   def auth_user_id
     request.headers['X-Govuk-Authenticated-User']
   end
 
+  def auth_organisation_id
+    request.headers['X-Govuk-Authenticated-User-Organisation']
+  end
+
   def auth_bypass_id
     request.headers['Govuk-Auth-Bypass-Id']
-  end
-
-  def not_access_limited(item)
-    !item.access_limited_by_user_id? && !item.access_limited_by_bypass_id?
-  end
-
-  def check_user_id(item)
-    item.user_id_allowed?(auth_user_id)
-  end
-
-  def check_bypass_id(item)
-    item.auth_bypass_id_allowed?(auth_bypass_id)
   end
 
   def json_forbidden_response
--- a/app/controllers/content_items_controller.rb
+++ b/app/controllers/content_items_controller.rb
@@ -64,24 +64,44 @@ private
     redirect_to route, status: 303
   end
 
+#  def can_view(item)
+#    auth_user_id != 'invalid' && (check_bypass_id(item) || check_user_id(item) || not_access_limited(item))
+#  end
+
   def can_view(item)
-    if auth_bypass_id_header.present?
-      item.includes_auth_bypass_id?(auth_bypass_id_header)
+    if valid_user?(auth_user_id)
+        if item.access_limited?
+          item.viewable_by_bypass_id?(auth_bypass_id) || item.viewable_by_user_id?(auth_user_id)
+        else
+          true
+        end
     else
-      !invalid_user_id? && item.viewable_by?(authenticated_user_uid)
+      false
     end
   end
 
-  def authenticated_user_uid
+  def valid_user?(user_id)
+    user_id != 'invalid'
+  end
+
+  def auth_user_id
     request.headers['X-Govuk-Authenticated-User']
   end
 
-  def auth_bypass_id_header
+  def auth_bypass_id
     request.headers['Govuk-Auth-Bypass-Id']
   end
 
-  def invalid_user_id?
-    authenticated_user_uid == 'invalid'
+  def not_access_limited(item)
+    !item.access_limited_by_user_id? && !item.access_limited_by_bypass_id?
+  end
+
+  def check_user_id(item)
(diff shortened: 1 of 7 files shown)


app/controllers/content_items_controller.rb
```diff
@@ -65,23 +65,43 @@ def redirect_canonical(content_item)
...
     else
-      !invalid_user_id? && item.viewable_by?(authenticated_user_uid)
+      false
     end
   end
 
-  def authenticated_user_uid
+  def valid_user?(user_id)
+    user_id != 'invalid'
+  end
```
kevindew: I think, but my understanding could be wrong, that a user would have an user_id of "invalid" if they are using an auth bypass token: https://github.com/alphagov/authenticating-proxy/blob/84b7b5d300fe7de8b328be964c421870a412bfd0/lib/proxy.rb#L95-L100

app/controllers/content_items_controller.rb
```diff
@@ -65,23 +65,43 @@ def redirect_canonical(content_item)
...
+      if item.access_limited?
+        item.viewable_by_bypass_id?(auth_bypass_id) || item.viewable_by_user_id?(auth_user_id) || item.viewable_by_org_id?(auth_org_id)
+      else
+        true
+      end
     else
-      !invalid_user_id? && item.viewable_by?(authenticated_user_uid)
+      false
     end
   end
```
kevindew: This seems to group the behaviour of auth_bypass into this, which I think would actually prevent people from having access.

I think you probably want the conditionals laid out like:

```
if auth bypass header is present
  check whether it's the one for this content
else if item has org or user access limit
  user is in user list or org is in org list
else
  true
end
```

does that make sense?

app/controllers/content_items_controller.rb
```diff
@@ -65,23 +65,43 @@ def redirect_canonical(content_item)
...
 
   def can_view(item)
```
kevindew: I feel like we should a question mark to this method since it only returns a bool

app/controllers/content_items_controller.rb
```diff
@@ -65,23 +65,43 @@ def redirect_canonical(content_item)
...
+
+  def auth_bypass_id
     request.headers['Govuk-Auth-Bypass-Id']
   end
 
-  def invalid_user_id?
-    authenticated_user_uid == 'invalid'
+  def not_access_limited(item)
+    !item.access_limited_by_user_id? && !item.access_limited_by_bypass_id?
+  end
```
kevindew: It doesn't look like anything uses this method

app/models/content_item.rb
```diff
@@ -171,12 +171,20 @@ def router_rendering_app
     rendering_app || "government-frontend"
   end
 
-  def viewable_by?(user_uid)
-    authorised_user_uids.empty? || authorised_user_uids.include?(user_uid)
+  def viewable_by_user_id?(user_id)
+    auth_user_ids.include?(user_id)
   end
```
kevindew: Thinking out load a bit but I wonder if we could encapsulate more of the viewing logic in the model rather than in the controller. Though that might be a bad idea if we have to return different responses based on different scenarios.

We could have a method:

```
def viewable?(auth_bypass_id, user_id, organisation_id)
  return auth_bypass_ids.include?(auth_bypass_id) if auth_bypass_id
  return true if auth_user_ids.empty? && auth_org_ids.empty?

  auth_user_ids.include?(user_id) || auth_org_ids.include?(organisation_id)
end
```

which could reduce the number of methods and logic scattered around, and provide a more concrete thing to test to make logic expectations clear.

app/models/content_item.rb
```diff
@@ -171,12 +171,20 @@ def router_rendering_app
...
+  def viewable_by_org_id?(org_id)
+    auth_org_ids.include?(org_id)
+  end
+
+  def viewable_by_bypass_id?(bypass_id)
+    auth_bypass_ids.include?(bypass_id)
+  end
+
+  def access_limited?
+    auth_bypass_ids.present? || auth_user_ids.present? || auth_org_ids.present?
```
kevindew: For this method I'd probably go with `auth_user_ids.any? || auth_organisation_ids.any?` and leave out auth_bypass_ids since that doesn't apply a limit (it's a bit confusing it's in that hash really)

app/models/content_item.rb
```diff
@@ -217,10 +225,14 @@ def should_register_routes?(previous_item: nil)
     true
   end
 
-  def authorised_user_uids
+  def auth_user_ids
     access_limited.fetch('users', [])
   end
 
+  def auth_org_ids
```
kevindew: I don't think you introduced this, so apologies. But I think it's premature to abbreviate this to org

app/controllers/content_items_controller.rb
theseanything: Yup - so having another look. Are we right in assume we don't need to do any check for `invalid` in the content store?

app/controllers/content_items_controller.rb
theseanything: Yup - we've had a go at writing the `viewable?`  method in the content item model.

app/models/content_item.rb
theseanything: Again yup - we've had a go at writing the viewable? method in the content item model.

app/models/content_item.rb
theseanything: Changed

app/models/content_item.rb
theseanything: Changed back to `organisation`

app/models/content_item.rb
```diff
@@ -171,12 +171,16 @@ def router_rendering_app
     rendering_app || "government-frontend"
   end
 
-  def viewable_by?(user_uid)
-    authorised_user_uids.empty? || authorised_user_uids.include?(user_uid)
+  def viewable?(bypass_id: nil, user_id: nil, organisation_id: nil)
+    return auth_bypass_ids.include?(bypass_id) if bypass_id
+    return user_id.present? if auth_bypass_ids.any?
```
kevindew: I don't think we need this line - when I first saw it it raised a concern to me of "what if user_id is 'invalid'" (this is what prompted me to suggest removing the "invalid" values in controller). Then thinking about it a bit more I don't think we care whether user is authenticated or not, as we expect that to be already handled by auth proxy.

app/controllers/content_items_controller.rb
```diff
@@ -64,24 +64,35 @@ def redirect_canonical(content_item)
...
+      user_id: auth_user_id,
+      organisation_id: auth_organisation_id
+    )
+
+    false
   end
 
-  def authenticated_user_uid
+  def auth_user_id
     request.headers['X-Govuk-Authenticated-User']
```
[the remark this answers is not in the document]
JonathanHallam: I think it would be better to remove the 'invalid' bit from the auth proxy than do the filter here. Do you know of any reason why we shouldn't?

app/models/content_item.rb
```diff
@@ -171,12 +171,16 @@ def router_rendering_app
...
+    return true if auth_user_ids.empty? && auth_organisation_ids.empty?
+
+    auth_user_ids.include?(user_id) || auth_organisation_ids.include?(organisation_id)
   end
 
-  def includes_auth_bypass_id?(auth_bypass_id)
-    auth_bypass_ids.include?(auth_bypass_id)
+  def access_limited?
```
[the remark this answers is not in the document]
JonathanHallam: The method is still called from another method that has to do with cache headers. I've left the method in but removed any calls from things that we've written for this.

PR #604, opened 2019-08-21
Use the `nearest` read mode for MongoDB

cbaines:
The Content Store is a read heavy application, as it only processes
writes when things change, but it's constantly serving the frontend
applications with data.

This can be seen in the load on the MongoDB machines, currently the
primary gets off lightly with a low load, and the secondary machines
handle most of the operations.

Using the nearest MongoDB machine should spread the load more evenly,
and might even increase performance due to reduced network latency.

I've been looking at this, as the MongoDB machines often complain
about high load (especially on the AWS side of Staging).

### DIFF
--- a/config/mongoid.yml
+++ b/config/mongoid.yml
@@ -31,4 +31,4 @@ production:
         write:
           w: <%= ENV['MONGO_WRITE_CONCERN'] || 'majority' %>
         read:
-          mode: :secondary_preferred
+          mode: :nearest


thomasleese:
Out of interest, did the load change after we improved the Publishing API to perform less dependency resolution? Because I imagine there was a significant drop in the number of writes after that.

cbaines:
> Out of interest, did the load change after we improved the Publishing API to perform less dependency resolution? Because I imagine there was a significant drop in the number of writes after that.

It's hard to tell, as I believe the primary machine handles writes, and the load on that is quite low. It doesn't appear to have changed recently though.

PR #629, opened 2019-10-10
Revert "Bump rspec-rails from 3.8.2 to 3.9.0"

ChrisBAshton:
This reverts commit 1cbad39c4d2bfb5264ddcce58d33d66738f55b69, as it started failing the build.


kevindew:
It might be better to pin sprockets to ‘~> 3’ rather than reverting rspec. This app has no use for sprockets and would isolate the problem.

ChrisBAshton:
Merging to unblock PRs. Sounds like the brakeman PR is being merged soon, so the next time we attempt to bump the version everything should work fine, without the overhead of having to remove our pinning.

PR #634, opened 2019-10-11
Use new auth bypass ids field

leenagupte:
Part of: Trello: https://trello.com/c/AhyYCLZy
Follows on from: PRs #630 and #631 
Related to:  PR #636 

# What's changed
Switch the code to use the new auth_bypass_ids field and no longer check or populate the old field

Uses the new auth_bypass_ids, and no longer populates it from the auth_bypass_ids in the access_limited field.

### DIFF
--- a/spec/factories/content_item.rb
+++ b/spec/factories/content_item.rb
@@ -20,6 +20,10 @@ FactoryBot.define do
       title { "" }
     end
 
+    trait :with_auth_bypass_id do
+      auth_bypass_ids { [SecureRandom.uuid] }
+    end
+
     factory :content_item do
       format { "answer" }
       title { "Test content" }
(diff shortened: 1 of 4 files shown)


spec/factories/content_item.rb
```diff
@@ -83,26 +83,22 @@
...
         }
       end
 
       trait :by_auth_bypass_id_and_org_id do
+        auth_bypass_ids { %w(85aa9fd5-c514-4964-b931-5b597e4ec668) }
         access_limited {
           {
-            "auth_bypass_ids" => %w(85aa9fd5-c514-4964-b931-5b597e4ec668),
             "organisations" => %w(f17250b0-7540-0131-f036-005056030202),
           }
```
kevindew: I'd suggest moving these traits to be separate from this `access_limited_content_item` factory.

I think it'd be be consistent to have a `with_auth_bypass_id` trait on the content item that defaults to a Random UUID. (similar to https://github.com/alphagov/content-store/blob/59a14cd8103340492638c3ea6494416824c8a10b/spec/factories/content_item.rb#L15-L17)

db/migrate/20191011142024_remove_auth_by_pass_i_ds_from_access_limited.rb
```diff
@@ -0,0 +1,10 @@
+class RemoveAuthByPassIDsFromAccessLimited < Mongoid::Migration
+  def self.up
+    ContentItem
+      .where(:access_limited.nin => [{}, nil])
+      .each { |ci| ci.unset("access_limited.auth_bypass_ids") }
```
kevindew: In my test this worked flawlessly 👍

db/migrate/20191011142024_remove_auth_by_pass_i_ds_from_access_limited.rb
```diff
@@ -0,0 +1,10 @@
...
+  end
+
+  def self.down
+  end
+end
```
kevindew: It might be an idea to hold fire on having this migration in here since this needs to occur after this code is running on production.

As it's a bit of a do or die moment we probably want to have it kept granular.

leenagupte: I'll cherry-pick this out into a separate PR.

leenagupte: https://github.com/alphagov/content-store/pull/636

PR #688, opened 2020-03-02
Use `rubocop-govuk` for linting instead of `govuk-lint`

issyl0:
- See commits for more details.
- https://trello.com/c/pR7GEFfU/128-make-all-current-govuk-repos-use-rubocop-govuk-instead-of-govuk-lint, part of the GOV.UK Developer Tooling group.


issyl0:
What does it want from me? This looks right:

```
lib/find_specific_term.rb:38:10: C: Rails/FilePath: Please use Rails.root.join('path/to') instead.
    puts "CSV file at #{Rails.root.join('/tmp/search_term_content_items.csv')}"
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

benjamineskola:
> What does it want from me? This looks right:
> 
> ```
> lib/find_specific_term.rb:38:10: C: Rails/FilePath: Please use Rails.root.join('path/to') instead.
>     puts "CSV file at #{Rails.root.join('/tmp/search_term_content_items.csv')}"
>          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> ```

If I remember correctly, it's an overzealous regex. It thinks that if you're doing `Rails.root` inside `#{}` it's because you're using it to do a join, i.e. it assumes it's `"#{Rails.root}/some/path"`. False positive, ignore. :)

issyl0:
I looked to fix it and it [looks like someone already tried](https://github.com/rubocop-hq/rubocop/commit/707c9a85c1d99bfee8a10ac404abf6a14d6a88f0) but obviously it hasn't stuck. 😕

benjamineskola:
I wonder if that's because they’ve only excluded using the multi-parameter form inside interpolations, and not the single-parameter-with-slashes form (the one that we use and is now actually the default). But I don't understand the matcher DSL they use well enough to follow it; I just remember it being an issue with some of the other apps a couple of months back.

issyl0:
This app uses MongoDB, not PostgreSQL. So it's Mongoid, not ActiveRecord. Which means that a lot of RuboCop's autocorrects won't work! Back to the start...

issyl0:
The fix to the Mongoid problem is to ignore the cops, for now: 0e50ea4.

PR #720, opened 2020-05-05
Add note about limitations of the content API

ChrisBAshton:
This lived in the Developer Docs until this PR:
https://github.com/alphagov/govuk-developer-docs/pull/2519

It will still display in the dev docs, but will be pulled
in from here, where it is closer to the code and can be
kept up to date more consistently.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,21 @@ Detailed technical information can be found in the
 Example API requests and corresponding responses can be found in the
 [content store pact-broker documentation][pact-broker-docs].
 
+## Limitations of the content API
+
+- Not all content is in the content-store yet. There are pages that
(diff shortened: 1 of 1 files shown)


thomasleese:
I'm not sure this makes sense in the README for the application, because from a perspective of the application it doesn't care what is in its database or not. I'm also not sure we even need this documentation anywhere as [it's written in our docs for the Content API](https://content-api.publishing.service.gov.uk/#what-you-can-39-t-do-with-this-api) - perhaps we need to make those docs more visible?

ChrisBAshton:
I'd forgotten those docs exist! Fair enough - will close this PR and look at ingesting those docs into the dev docs (or at least signposting them more clearly).

PR #728, opened 2020-05-26
Typo in docs. Inconsistency between actual and stated number of contexts

reggieb:
The text in `doc/content_item_fields.md` states:

> The three contexts are:

But only two contexts are listed. This PR changes that text to:

> The two contexts are:

### DIFF
--- a/doc/content_item_fields.md
+++ b/doc/content_item_fields.md
@@ -1,7 +1,7 @@
 # The Content Item Format
 
 A content item consists of a set key/value pairs. Different key/value pairs are
-present or required, depending on the context. The three contexts are:
+present or required, depending on the context. The two contexts are:
 
  - storing: content items being sent to the content store.
  - retrieving: content items being retrieved from the content store.


reggieb:
[Later in the document](https://github.com/alphagov/content-store/blob/master/doc/content_item_fields.md#representation-of-content-item-links-in-different-contexts) a `notifying` context is mentioned. Is this an error or should there in fact be three contexts listed!

kevindew:
Thanks for taking this on 👍 

Unfortunately this doc seems super out of date with how the content store works 🙈 Since about https://github.com/alphagov/content-store/pull/231 which is alarmingly nearly 4 years ago. There's now no longer these "storing", "notifying" and "retrieving" contexts. Apologies, that you may feel like you've opened up a huge can of worms here.

My simple suggestions to try get things to a degree of up-to-date would be to remove the `input_examples` and `output_examples` directories. I doubt there has been much motivation to keep those up to date. Then for this content item page, there is probably on-going risk that bringing this up-to date will have it slip out of date again so it might be better to point to https://docs.publishing.service.gov.uk/content-schemas.html which is auto generated based on the schemas.

reggieb:
OK - I'll put this on the back burner, and use it as something I can work on between tickets.

brucebolt:
Closing this ticket as it hasn't been commented on for a couple of years.

PR #743, opened 2020-06-24
Sync this repo with RuboCop GOV.UK

benthorner:
https://github.com/alphagov/rubocop-govuk/issues/73

Please see the commits for more details.

### DIFF
--- a/spec/integration/fetching_content_item_spec.rb
+++ b/spec/integration/fetching_content_item_spec.rb
@@ -126,7 +126,7 @@ describe "Fetching content items", type: :request do
   end
 
   context "a content item with a non-ASCII base_path" do
-    let(:content_item) { create(:content_item, base_path: URI.encode("/news/בוט לאינד")) }
+    let(:content_item) { create(:content_item, base_path: Addressable::URI.encode("/news/בוט לאינד")) }
 
     before(:each) { get "/content#{content_item.base_path}" }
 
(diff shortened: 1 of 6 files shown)


spec/integration/fetching_content_item_spec.rb
```diff
@@ -126,7 +126,7 @@
   end
 
   context "a content item with a non-ASCII base_path" do
-    let(:content_item) { create(:content_item, base_path: URI.encode("/news/בוט לאינד")) }
```
thomasleese, line 129: Thinking about these tests I'm not sure they're really testing the behaviour if we're just calling `Addressable::URI.encode`, it feels like we should hard code the URL encoded string. This would also allow us to find any potential problems in changing from `URI.escape` to `Addressable::URI.encode`.

benthorner: Sure, I think I understand your concern. I've added an extra commit - can you take another look?

thomasleese: That seems like a strange rule about ASCII comments... but okay.

benthorner: Indeed. However, this is one of the very few instances where we have such characters. The advantage of this Cop is it stops people entering weird whitespace, (fancy) quote and even mathematical characters (in Smart Answers), which are all difficult to reproduce.

kevindew: Hold up - is that going to prevent emojis? 😿

PR #744, opened 2020-07-06
Enable migrations to run against a new, empty cluster

bilbof:
When we tried to deploy content store in a new environment, with a fresh mongodb, these migrations failed.

This adds fixes for the various failures so we can run migrations against a new empty mongodb.


bilbof:
Thanks @kevindew, that makes sense. They are mostly data migrations as you say. I found one of that failed did try to create/delete a collection: https://github.com/alphagov/content-store/pull/744/files#diff-624484f797ed66f4da37981f330445bcR3.

I'll merge this in and put up another PR for removing the db/migrate directory.

PR #745, opened 2020-07-06
Remove db/migrations

bilbof:
Following from @kevindew's suggestion, this removes the db migrations from content-store:
https://github.com/alphagov/content-store/pull/744#pullrequestreview-442966860

This is to discourage having 'data migrations' and instead have developers introduce temporary rake tasks for one-off tasks that they need to perform.


kevindew:
Nice one @bilbof - looks good to me. Good to say goodbye to a lot of code. It'd be good to also remove the gem too: https://github.com/alphagov/content-store/blob/2d5dc852291f911f45ff5264fa782699c840483d/Gemfile#L9

bilbof:
Thanks! I've added a couple more commits.

PR #746, opened 2020-07-07
Upgrade rails to 6.0.3

NeilvB:


thehenster:
Suggested fix for the build failure: https://github.com/thehenster/content-store/commit/6e6caa6febb6e041ae3509cdb115017bba317464

It looks to me like `config/puma.rb` was added by accident once upon a time. It looks like unicorn is the server that's actually used. I suppose we can delete `config/puma.rb`.

`config/cable.yml` can be deleted.

`script` folder can be deleted. It's a relic.

config/environments/production.rb
```diff
@@ -14,36 +14,21 @@
...
+  # Ensures that a master key has been made available in either ENV["RAILS_MASTER_KEY"]
+  # or in config/master.key. This key is used to decrypt credentials (and other encrypted files).
+  # config.require_master_key = true
 
   # Disable serving static files from the `/public` folder by default since
   # Apache or NGINX already handles this.
   config.public_file_server.enabled = ENV["RAILS_SERVE_STATIC_FILES"].present?
 
-  # Compress JavaScripts and CSS.
-  # config.assets.js_compressor = :uglifier
```
EGiataganas, line 27: Can you please revert these 2 lines and enable the js_compressor option for Sprockets?

```ruby
  # Compress JS using a preprocessor.
  config.assets.js_compressor = :uglifier
```

MuriloDalRi (approved):

PR #750, opened 2020-07-13
Add Dependabot configuration

ChrisBAshton:
Add Dependabot configuration

In accordance with [RFC 126], we're configuring Dependabot to only
raise Pull Requests for the following three categories:

1. Security updates
2. Internal libraries
3. Framework libraries

Things have moved on slightly since the RFC; we're now using
[GitHub native] configs rather than the legacy `.dependabot`
approach. The syntax is slightly different to the example config
provided in the RFC but most of it is easily transferable.

We've handled (2) and (3) in config, and have set it to only raise
PRs for direct updates, i.e. updates to subdependencies of the
named dependencies will not raise PRs.

We've chosen `gds*` and `govuk*` wildcards to cover things like
`gds-api-adapters` and `govuk_app_config`. We could have allowed
these as suffixes too (`*govuk`) but that naming convention is
not encouraged, so instead we've named `rubocop-govuk` explicitly.
[Read more about our wildcard choices][wildcards].

Trello: https://trello.com/c/uPoriyfJ/2049-add-dependabot-configuration-to-each-repo-blitz-pair

[GitHub native]: https://docs.github.com/en/github/administering-a-repository/enabling-and-disabling-version-updates
[security]: https://github.community/t/how-to-get-dependabot-to-trigger-for-security-updates-only/117257/5
[RFC 126]: https://github.com/alphagov/govuk-rfcs/blob/master/rfc-126-custom-configuration-for-dependabot.md
[wildcards]: https://github.com/alphagov/content-store/pull/750#discussion_r453693082

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,34 @@
+version: 2
+updates:
+  - package-ecosystem: bundler
+    directory: /
+    schedule:
+      interval: daily
+    allow:
+      # Security updates
+      - dependency-name: brakeman
+        dependency-type: direct
+      # Internal gems
+      - dependency-name: "gds*"
+        dependency-type: direct
+      - dependency-name: "govuk*"
+        dependency-type: direct
+      - dependency-name: "rubocop-govuk"
+        dependency-type: direct
+      - dependency-name: plek
+        dependency-type: direct
+      # Framework gems
+      - dependency-name: factory_bot
+        dependency-type: direct
+      - dependency-name: rails
+        dependency-type: direct
+      - dependency-name: rspec-rails
+        dependency-type: direct
+      - dependency-name: mongoid
+        dependency-name: direct
+  - package-ecosystem: docker
+    directory: /
+    schedule:
+      interval: weekly
+    ignore:
+      - dependency-name: ruby


.github/dependabot.yml
```diff
@@ -0,0 +1,34 @@
...
+      # Security updates
+      - dependency-name: brakeman
+        dependency-type: direct
+      # Internal gems
+      - dependency-name: "gds*"
+        dependency-type: direct
+      - dependency-name: "gov*"
+        dependency-type: direct
+      - dependency-name: "*-govuk"
+        dependency-type: direct
```
kevindew, line 17: Exciting that this has been raised 🎉 

I feel like the gds* and *-govuk ones are a bit cryptic. I think we'd only need a wildcard on govuk. Have we got enough consistency in naming that we can do govuk_* everywhere or do we have govuk-* ones lingering away? (they're naughty)

ChrisBAshton: Thanks for the comment, Kevin (I edited it just to escape some of the markdown GitHub was rendering)

These wildcards are based on the example config in the [RFC](https://github.com/alphagov/govuk-rfcs/blob/master/rfc-126-custom-configuration-for-dependabot.md):

```
      - match: { dependency_name: gds-api-adapters }
      - match: { dependency_name: gds-sso }
      - match: { dependency_name: govspeak }
      - match: { dependency_name: govuk* }
      - match: { dependency_name: rubocop-govuk }
      - match: { dependency_name: plek }
      - match: { dependency_name: scss_lint-govuk }
```

It looks like we need both `gov*` and `*-govuk`:

* `gov*` catches `govspeak`, `govuk-content-schemas`, etc
* `*-govuk` catches `rubocop-govuk`, `scss_lint-govuk` etc.

I could consider combining into one single `*gov*` wildcard, although this risks including non-GOVUK dependencies.

kevindew: Sure, I'd simplify it to have `govuk*` and have `govspeak`, `rubocop-govuk` and `scss_lint-govuk` be explicit ones. We shouldn't have too much to gain from `*-govuk` anyway as we're actively working towards replacing `scss_lint-govuk` so that'll only leave rubocop-govuk.

If anything only having the one govuk* one helps encourage people that `govuk_*` is the conventional way to name a govuk gem and other ways are exceptional.

ChrisBAshton: Done 👍 hope you don't mind me linking to this conversation from the commit.

kevindew: Thanks @ChrisBAshton, the explicit govuk* ones look good.

Sorry to be a pedant but something we didn't resolve in this conversation was the gds* ones. I don't think they should be done with an asterisk as these are non-standard names for GOV.UK gems. There shouldn't be new ones created so it seems reasonable the two we have are explicitly referenced. I'm just being a pain on this as I imagine this will start spreading across the stack and will be harder to resolve later.

PR #757, opened 2020-09-02
Make find_by_path more efficient

kevindew:
This makes a couple of changes to the FindByPath module to try improve its performance. Firstly it only checks redirects if there are any, this allows PublishIntent to make it's queries by index scans rather than collection scans. Secondly it splits the work into two queries, the first being a simple find by id and only if that fails doing the query to check the routes/redirects.

I'm going to run this on integration/staging for a bit to try determine if it has an impact.

In integration:

Screenshot 2020-09-02 at 10 18 57

In staging:

Screenshot 2020-09-02 at 10 20 00

It seems that this makes a noticeable impact on PublishIntent, yet only a negligible impact on ContentItem look ups. It also doesn't seem to fully remove spikes as this later graph showed:

Screenshot 2020-09-02 at 10 24 03

So this may well help the situations that have caused problems, but it looks like it won't fully resolve them.

### DIFF
--- a/lib/find_by_path.rb
+++ b/lib/find_by_path.rb
@@ -13,15 +13,17 @@ class FindByPath
   end
 
   def find(path)
-    matches = find_matching_items(path)
-    matches.present? ? best_match(matches, path) : nil
+    exact_match = model_class.where(base_path: path).find_first
+    return exact_match if exact_match
+
+    matches = find_route_matches(path)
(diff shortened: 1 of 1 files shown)


cbaines (approved):
Sounds good to me :+1:

Tetrino (approved):

PR #758, opened 2020-09-02
Fix flakey scheduled publishing log test

kevindew:
This test seemed to rely on the code running within 0.01 seconds which
is variable based on a number of factors. I've switched to use Timecop
to remove this environmental impact.


cbaines (approved):
Sounds good to me :+1:

PR #790, opened 2020-12-11
[WIP] Spike adding /api/content-id/:content_id route

ChrisBAshton:
This work explores adding a new public endpoint we can call to
quickly find out what content is represented by a specific content
ID. In absence of a route, the alternative has been to connect to
the VPN, SSH into a Content Store or Publishing API machine and
perform a lookup via the app console.

This idea isn't particularly new. We have a precedent in Whitehall:
https://github.com/alphagov/whitehall/blob/0ef9b6bb99e62e1e7d724b2cd9c7da4233d1bca4/config/routes.rb#L191

This spike works locally. If you app console and compare
`app.get("/api/content")` (the route of the homepage) with
`app.get("/api/content-id/f3bbdec2-0e62-4520-a7fd-6ffd5d36e03a")`
(the content ID of the homepage), you get the same `app.response`.

To make this a viable MVP, we'd need to [add an index](https://api.rubyonrails.org/v6.1.0/classes/ActiveRecord/ConnectionAdapters/SchemaStatements.html#method-i-add_index) on the `content_id`,
as currently the lookup is VERY slow (1-2 minutes per query!).

(We should **consider doing this anyway**, as it's a useful thing to be able
to query and currently queries are just too slow.)

And then to make this available via a public endpoint
rather than app console, we'd need to add a 'special route' like
we've done for `/api/content`:
https://github.com/alphagov/special-route-publisher/blob/c47ca1b995315d5abf3c811a321c7520467fe98e/data/special_routes.yaml#L1-L6

### DIFF
--- a/app/controllers/content_items_controller.rb
+++ b/app/controllers/content_items_controller.rb
@@ -1,5 +1,5 @@
 class ContentItemsController < ApplicationController
-  skip_before_action :authenticate_user!, only: [:show]
+  skip_before_action :authenticate_user!, only: %i[show show_by_id]
   before_action :parse_json_request, only: [:update]
 
   def show
@@ -15,6 +15,29 @@ class ContentItemsController < ApplicationController
 
     return error_404 unless item
     return redirect_canonical(item) if item.base_path != encoded_request_path
+
+    if can_view?(item)
+      render json: ContentItemPresenter.new(item, api_url_method), status: http_status(item)
+    else
+      render json_forbidden_response
+    end
+  end
+
+  def show_by_id
+    item = GovukStatsd.time("show.find_content_item") do
+      ContentItem.find_by(content_id: params[:content_id])
+    end
+
+    intent = nil
+    if item
+      intent = GovukStatsd.time("show.find_publish_intent") do
+        PublishIntent.find_by_path(item.base_path)
+      end
+    end
+
+    set_cache_headers(item, intent)
+
+    return error_404 unless item
 
     if can_view?(item)
       render json: ContentItemPresenter.new(item, api_url_method), status: http_status(item)
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -2,6 +2,7 @@ Rails.application.routes.draw do
   scope format: false do
     # The /api/content route is used for requests via the public API
     get "/api/content(/*path_without_root)" => "content_items#show", :as => :content_item_api, :public_api_request => true
+    get "/api/content-id/:content_id" => "content_items#show_by_id"
 
     get "/content(/*path_without_root)" => "content_items#show", :as => :content_item
     put "/content(/*base_path_without_root)" => "content_items#update"


kevindew:
Nice idea Chris, here comes a quick fire some rapid unsolicited comments you didn't anticipate because I saw this pop up in my mailbox 😆 .

First I wondered, do you have a concrete example of the use case for this? that might be useful to explain the advantages of this proposal. I didn't know what scenarios we expected someone to have a content id and needed the item.

The first thing that came to mind seeing this is that content id is not a unique identifier in Content Store since we can have multiple content items with the same content id but differing locales.

The other more nitpicking reactions I had was that the url `/api/content-id/:id` doesn't feel a very restful URL. It sounds like you are looking up a content id rather than a content item. It also doesn't feel great to have the same code as in show and show_by_id. My quick fire thought was that it might be better to have `/content/?content_id=:id` with an optional locale parameter that defaults to en, this could then determine the content item and redirect you to `/api/content/:base_path` - but I'm not sure how restful that feels either - and thinking about it that would conflict with the homepage route hmm.

Also, I'm assuming you're wanting this for public usage - i.e an equivalent of `https://www.gov.uk/api/content/vat-rates`. This opens up some frustrating semantic questions. The prefix of `/api/content` is essentially the [base URL of the public Content API](https://content-api.publishing.service.gov.uk/reference.html#gov-uk-content-api-v1-0-0) which makes something like `https://www.gov.uk/api/content-id/:id` somewhat strange as it has jumped beyond the bounds of the Content API and is now a new rogue API. Of course, with [the `*` routing](https://github.com/alphagov/content-store/blob/master/config/routes.rb#L4), it's virtually impossible to extend this API with a more specific path. 

A fallback idea that comes to mind (that is pretty hacky) is that you could have `/api/content/:locale/:content_id` which uses the same route as `/api/content/:base_path` but only looks up the item if there is no matching route, so would work on the expectation that for `/vat-rates` there is virtually no risk that a different page has the base_path of `/en/f838c22a-b2aa-49be-bd95-153f593293a3`. Can't say I'm particularly enthused about that idea though.

That rounds up my thoughts, I feel like I've mostly achieved the effect of opening 5 different cans of worms on the idea. Hope that helps 😬 .  You can consider a parallel universe where we implemented v2 of the Content API where this was designed in from the start: https://github.com/alphagov/govuk-rfcs/blob/master/rfc-081-approach-for-next-iteration-of-content-api.md#documentscontent_id

ChrisBAshton:
Thanks for the review, @kevindew - I expected nothing less! 😆

> First I wondered, do you have a concrete example of the use case for this? that might be useful to explain the advantages of this proposal. I didn't know what scenarios we expected someone to have a content id and needed the item.

I had one today where I wanted to find out more about a piece of content referenced [in a PR](https://github.com/alphagov/govuk_publishing_components/pull/1821). And often when I browse through errors in Sentry (such as [this one](https://sentry.io/organizations/govuk/issues/1970324283/?project=202242&query=is%3Aunresolved&statsPeriod=24h)), all I have to go on is a content ID. Whilst I know how to find out more info via VPN -> SSH -> Console -> Query, I often just want to quickly check what I'm looking at so I can write up a comment about it and deal with it later.

> The first thing that came to mind seeing this is that content id is not a unique identifier in Content Store since we can have multiple content items with the same content id but differing locales.

Very good point. This is just a spike, but I envisage any production implementation to have two parameters, so might look more like `/api/content-id/:content_id/:locale`.

> The other more nitpicking reactions I had was that the url `/api/content-id/:id` doesn't feel a very restful URL. It sounds like you are looking up a content id rather than a content item.

Agreed. I'd want to get consensus before committing to a URL structure. For the purposes of the spike I wanted something that avoids the [wildcard catch all](https://github.com/alphagov/content-store/blob/master/config/routes.rb#L4).

> It also doesn't feel great to have the same code as in show and show_by_id. My quick fire thought was that it might be better to have `/content/?content_id=:id` with an optional locale parameter that defaults to en, this could then determine the content item and redirect you to `/api/content/:base_path` - but I'm not sure how restful that feels either - and thinking about it that would conflict with the homepage route hmm.

I like this idea. Why would this not work for the homepage? Would it not just redirect you to the same page but without the URL query parameters? It seems neater than the other proposed ideas.

kevindew:
> I had one today where I wanted to find out more about a piece of content referenced [in a PR](https://github.com/alphagov/govuk_publishing_components/pull/1821). And often when I browse through errors in Sentry (such as [this one](https://sentry.io/organizations/govuk/issues/1970324283/?project=202242&query=is%3Aunresolved&statsPeriod=24h)), all I have to go on is a content ID. Whilst I know how to find out more info via VPN -> SSH -> Console -> Query, I often just want to quickly check what I'm looking at so I can write up a comment about it and deal with it later.

Sure, so it's essentially a developer debugging mechanism?

> I like this idea. Why would this not work for the homepage? Would it not just redirect you to the same page but without the URL query parameters? It seems neater than the other proposed ideas.

The problem we've got is that `https://www.gov.uk/api/content` is the resource of the homepage. So if we had `https://www.gov.uk/api/content?content_id=:id` then that would be passing query strings to the homepage content item (like so: https://www.gov.uk/api/content?content_id=f838c22a-b2aa-49be-bd95-153f593293a3&locale=en) rather than a means of specifying a distinct resource. In a more restful world you'd consider `/api/content/` to be some form of index that could be queried and `/api/content/:base_path` to be a show resource.

ChrisBAshton:
> Sure, so it's essentially a developer debugging mechanism?

Yes. Much like the [precedent we have in Whitehall](https://github.com/alphagov/whitehall/blob/0ef9b6bb99e62e1e7d724b2cd9c7da4233d1bca4/config/routes.rb#L191) - a way of performing a task more quickly and easily than the alternatives we currently have available.

> The problem we've got is that `https://www.gov.uk/api/content` is the resource of the homepage. So if we had `https://www.gov.uk/api/content?content_id=:id` then that would be passing query strings to the homepage content item (like so: https://www.gov.uk/api/content?content_id=f838c22a-b2aa-49be-bd95-153f593293a3&locale=en) rather than a means of specifying a distinct resource. In a more restful world you'd consider `/api/content/` to be some form of index that could be queried and `/api/content/:base_path` to be a show resource.

I agree it's not very RESTful, but I think this would still work fine. I imagine Content API looking for the presence of a `content_id` parameter first and foremost, and only looking at the route in the URL if no parameter exists. So in the homepage example, a visit to https://www.gov.uk/api/content?content_id=f838c22a-b2aa-49be-bd95-153f593293a3&locale=en would perform a lookup on the content ID, find the base path is "/", and redirect to https://www.gov.uk/api/content.

kevindew:
I'm going to close this since it is now quite stale, fine to re-open if/when not stale.

PR #822, opened 2021-04-12
Correct port number

ChrisBAshton:
Looking at the commit where this was added (d11e9e9cbb97a7ec9729d637535d6b4e2fc96950),
the port number should actually have been `3068` to match the
value in the (then) startup.sh file and the (now) Dockerfile:
https://github.com/alphagov/content-store/blame/bbd42092fce84754c084f0c496240ec0741df34e/Dockerfile#L8

`3086` seems to be being used by the Transition app:
https://github.com/search?q=org%3Aalphagov+3086&type=code

This was set 3 years ago, however, so begs the question: what has
the impact of this been and why have we not noticed?

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,1 +1,1 @@
-web: bundle exec unicorn -c ./config/unicorn.rb -p ${PORT:-3086}
+web: bundle exec unicorn -c ./config/unicorn.rb -p ${PORT:-3068}


kevindew (approved):
Nice catch, I imagine the impact of this has been negligible

This procfile will have only been regularly used by publishing-e2e-tests by way of the [Dockerfile](https://github.com/alphagov/content-store/blob/a7ba83421ac5394c01970ee96eb8cef345561fd1/Dockerfile#L19) - and the PORT env var is set directly in [docker-compose.yml](https://github.com/alphagov/publishing-e2e-tests/blob/526b4b2edd88a6eeb6634d35c7186e51f2b265ca/docker-compose.yml#L289) meaning this fallback won't have taken affect unless someone ran it manually.

PR #823, opened 2021-04-13
Remove `pact:verify:branch[BRANCH_NAME]` rake task

ChrisBAshton:
Depends on https://github.com/alphagov/publishing-api/pull/1934. Proof of passing build: https://ci.integration.publishing.service.gov.uk/job/content-store/job/simplify-pact-verify/8/console

We don't need to expose an additional task to test against a branch of the consumer. This also renames 'PUBLISHING_API_PACT_BRANCH' => 'PACT_CONSUMER_VERSION', as implemented in alphagov/gds-api-adapters#1049.

See commits for details.

Trello: https://trello.com/c/rnfUFP5o/2453-experiment-with-common-configuration-for-pact-tests-timebox-2-days


thomasleese (approved):

kevindew:
Quick comment as I imagine you're about to merge....

I don't think your last commit is quite right here. There is a deployed-to-production pact: https://pact-broker.cloudapps.digital/pacts/provider/Content%20Store/consumer/Publishing%20API/versions/branch-deployed-to-production

As I understand the code used to automatically prefix deployed-to-production with `branch-`. If i remember correctly it was preferable to test against production deployed version to prevent merging code that breaks prod, however this can't really be done with a gem (GDS API Adapters) which is why that uses master.

ChrisBAshton:
Thanks @kevindew, I'd missed the `branch-` prefix for `deployed-to-production`. Will rewrite that commit 👍

PR #826, opened 2021-04-20
Switch read mode to secondaryPreferred

bilbof:
We are seeing content-store read requests time out.

This change to the read prefence [1] is intended to
see if we can reduce reads to the primary, which seems
to have a higher percentage rate of locks.

[1] https://docs.mongodb.com/v2.6/core/read-preference/


kevindew (approved):

smford (approved):
LGTM, logic is sound

PR #827, opened 2021-04-22
Bundle update mongoid

kevindew:
This is a quick proof of concept to see if we can upgrade Mongoid even
on an app that uses Mongo 2.6.

It then fixes the error that stopped mongoid being updated as there was a typo.


.github/dependabot.yml
```diff
@@ -27,7 +27,7 @@ updates:
       - dependency-name: rspec-rails
         dependency-type: direct
       - dependency-name: mongoid
-        dependency-name: direct
```
ChrisBAshton, line 30: 😱 thanks for catching this!

ChrisBAshton (approved):
Thanks for your help, Kevin. I think this is good to go, even as a POC, right?
I've shared your findings on the https://trello.com/c/V56Gndqj/2216-investigate-fix-dependabot-not-raising-rails-61-prs Trello card.

PR #872, opened 2021-10-05
Adding support for publishing images to ECR using Github actions

nsabri1:
https://trello.com/c/mCSVqNIZ/642-implement-github-actions-as-ci-pipeline

### Github Actions CI:
A new Github workflow has been added, which will publish images to AWS ECR, each image will be tagged with Github sha.

Images will be built on both PR and merge to Main.

The build relies on a Github organisational secret being accessible to this repo:
    AWS_GOVUK_ECR_ACCESS_KEY_ID
    AWS_GOVUK_ECR_SECRET_ACCESS_KEY


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,101 @@
+name: Publish to ECR
```
kevindew, line 1: This name seems a little misleading if this also does testing.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,101 @@
...
+      - "Jenkinsfile"
+      - ".git**"
+
+  pull_request:
+    paths-ignore:
+      - "helm/**"
+      - "**.md"
+      - "**.txt"
+      - "Jenkinsfile"
+      - ".git**"
```
kevindew, line 20: This seems perhaps too defensive to me. Are we expecting problems if we have the occasional identical build? It feels unlikely we'd ever get an exhaustive list of things here and potentially a bit dangerous if applied uniformly(some apps may use a markdown or text file)

It also has the practical problem that we can't see if changes to the workflow script do anything.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,101 @@
...
+
+jobs:
+  test:
```
kevindew, line 23: It seems a bit of a pain that this will mean the project is tested twice (once via Jenkins again by this). It might be better to scope this just to pushing to ECR and not doing testing here?

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,101 @@
...
+        with:
+          path: node_modules
+          key: yarn-${{ hashFiles('**/yarn.lock') }}
+          restore-keys: yarn
+
+      - name: Install JavaScript dependencies
+        run: yarn install --frozen-lockfile
+
+      - name: Precompile assets
+        run: bundle exec rails assets:precompile
```
kevindew, line 66: Minor: this app doesn't have assets or yarn dependencies.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,101 @@
...
+
+      - name: Retrieve cached Gems
+        uses: actions/cache@v2
+        with:
+          path: vendor/bundle
+          key: bundle-v2-${{ hashFiles('**/Gemfile.lock') }}
+          restore-keys: bundle-v2
+
+      - name: Install Ruby dependencies
+        run: bundle install --jobs 4 --retry 3 --deployment
```
kevindew, line 50: Minor, this can be done a bit more elegantly with:

```
      - uses: ruby/setup-ruby@v1
        with:
          bundler-cache: true
```

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,101 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          aws-access-key-id: ${{ secrets.AWS_ECR_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_ECR_SECRET_ACCESS_KEY }}
```
kevindew, line 83: Since these are org level (GDS) secrets do we need to prefix them to indicate that they are GOVUK specific? I assume there's other programmes in GDS that may also want an ECR secret and it'd be good if it's clear this one belongs to GOV.UK (that's me assuming that we're using separate ECR repos)

helm/intergration/values.yaml
```diff
@@ -0,0 +1,57 @@
+# Default values for content-store.
```
kevindew, line 1: I can see from your description about where these files belong. My suggestion is that they don't belong in this repo as there's nothing to test them or any interaction with the app code.

I imagine that if we need to change something across lots of the helm files (such as change a hostname) it'd be a much simpler job for one repo with them all compared to PRs in each repo.

kevindew (commented):
Hi @nsabri1, I was a bit surprised to be asked for this review as this seems perhaps something your team members have more opinions on than me but felt like I could offer a few comments from a GOV.UK developer perspective. I hope they're helpful.

First, nice job with the action file. The pushing to ECR looks clean.

There is a bit of a divisive comment I'm making. Have you considered using Jenkins as the means to push to ECR? We already have code in the [Jenkins groovy script](https://github.com/alphagov/govuk-jenkinslib/blob/e69df4d1caaf5a9039af8a44fd4fbfe3f9b9a813/vars/govuk.groovy#L273-L281) to push to two repos of Docker images (we actively push to DockerHub now) and so it seems only a small stretch to add a third.

While I know that we wish to move away from Jenkins in the future it seems that we're going to end up duplicating a bunch of stuff if we have some things running in Jenkins and some in GitHub Actions that'll increase complication. Whereas to do this in Jenkins would be a bit cleaner and could just be ported to a GitHub action when we can retire Jenkins.

Anyway, that seems an idea to discuss with your team further.

.github/workflows/ci.yaml
nsabri1: i was actually suppose to name this Test, build and publish to ECR ! will update this

PR #876, opened 2021-10-19
Add Puma via govuk_app_config

sophiebrown01:
This PR adds `Puma` via govuk_app_config. Puma was previously installed indirectly through the dependency on `govuk_test` as part of the `dev` and `test` groups; given that we want to run Puma in production, we are instead configuring it via govuk_app_config.

**This PR is reliant on [changes to govuk_app_config](https://github.com/alphagov/govuk_app_config/pull/214)**

[Trello card](https://trello.com/c/irrbEFMV/668)


karlbaker02:
To expand on this, I don't think we want to make the switch to use Puma yet in the same way that Unicorn is defined (by using `puma.rb` to call for example a new `GovukPuma` defined in govuk_app_config), because we don't want to switch over apps in the existing stack to use Puma. The idea is that as part of the replatforming work, we will switch over to Puma when the apps are run on K8s, so to facilitate this we're changing the app Dockerfiles to explicitly state this as the `CMD` (equally it can be defined in K8s deployments).

Once we're running on K8s and the old stack has been deprecated, it should be fairly simple to take these configs and standardise them by setting them in govuk_app_config.

@sengi do you agree with this assessment?

kevindew:
> To expand on this, I don't think we want to make the switch to use Puma yet in the same way that Unicorn is defined (by using `puma.rb` to call for example a new `GovukPuma` defined in govuk_app_config), because we don't want to switch over apps in the existing stack to use Puma. The idea is that as part of the replatforming work, we will switch over to Puma when the apps are run on K8s, so to facilitate this we're changing the app Dockerfiles to explicitly state this as the `CMD` (equally it can be defined in K8s deployments).
> 
> Once we're running on K8s and the old stack has been deprecated, it should be fairly simple to take these configs and standardise them by setting them in govuk_app_config.
> 
> @sengi do you agree with this assessment?

One thing to be aware of is that setting anything up in govuk_app_config won't mean anything changes automatically for our apps. They're start a [specific unicorn process](https://github.com/alphagov/govuk-puppet/blob/32c1bbbb10067078c1406170666a135b4a10aaea/modules/govuk/files/usr/local/bin/govuk_spinup#L40), so won't switch over. You won't need to wait for the old stack to be deprecated before adding things there  - if anything, it sounds like a lot of admin to update every app first and then move it to govuk_app_config - and it's pretty low cost as nearly all the apps already have puma as an indirect dependency through govuk_test

PR #883, opened 2021-11-24
Reduce default TTL for GOV.UK pages to 20 minutes

bilbof:
Partially implements GOV.UK RFC 144. We will gradually reduce the TTL to 5 minutes, once we're more confident on the impact on performance. The next step will be to audit frontend applications to ensure that this TTL is honoured.

To be merged once https://github.com/alphagov/govuk-rfcs/pull/144 is approved.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -106,7 +106,7 @@ module ContentStore
     ]
 
     # Caching defaults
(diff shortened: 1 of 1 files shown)


kentsanggds (approved):
LGTM

PR #885, opened 2021-11-30
Set RAILS_LOG_TO_STDOUT in Dockerfile.

sengi:
This always needs to be set when a Rails app is running in a container, so it makes sense to set it in the Dockerfile rather than making the user remember to set it.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,7 +16,7 @@ RUN bundle config set without 'development test' && \
 COPY . ./
 
 FROM $base_image
-ENV RAILS_ENV=production GOVUK_APP_NAME=content-store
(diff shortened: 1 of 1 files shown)


kevindew:
You might find that this is the start of a yak shave I'm afraid. We do some whacky things with redirecting logs from STDOUT to STDERR in a gem: https://github.com/alphagov/govuk_app_config/blob/main/lib/govuk_app_config/govuk_logging.rb 😬

sengi:
Turns out this isn't needed; see alphagov/frontend#3055.

sengi:
Thanks, Kevin - I'd just found that and realised before seeing your comment there! At least we're already getting the logs to where they need to be. The buffering issue might be fun to fix though 😅 (I'm hoping it's just a Puma bug that's already fixed in a newer version but we'll see)

PR #897, opened 2022-02-01
Upgrade Ruby to 2.7.5

ChrisBAshton:
Upgrades Ruby to 2.7.5. See commits for more details.
            This PR should be tested before merge, as per https://docs.publishing.service.gov.uk/manual/ruby.html#update-ruby-version-in-the-relevant-repos.


Dockerfile
```diff
@@ -1,4 +1,4 @@
-# TODO: make this default to govuk-ruby once it's being pushed somewhere public
+FROM ruby:2.7.5
```
davidgisbey, line 1: I had the same thing. This line needs to come out

Dockerfile
```diff
@@ -1,4 +1,4 @@
...
 # (unless we decide to use Bitnami instead)
 ARG base_image=ruby:2.7.2-slim-buster
```
davidgisbey, line 3: And this needs to be updated to `ARG base_image=ruby:2.7.5-slim-buster`

These PRs were automated. It works when the 1st line of the file is `FROM ruby:2.7.2` as it replaces it with `FROM ruby:2.7.5`, but when a variable is set and reused throughout the dockerfile it breaks it.

beccapearce (approved):

PR #926, opened 2022-04-08
Update to Ruby 3.0.3

kevindew:
Amazingly nothing to fix or gems to update.


BeckaL (approved):
Wow this feels almost suspiciously easy...

PR #946, opened 2022-06-28
Handle errors for bad URI requests

kevindew:
We occasionally are subject to malicious users trying to discover
weaknesses in our application by sending invalid URIs.

This change updates our handling so that we return a 400 Bad Request
response, rather than recording this as an application error.

I debated about whether to return a 400 (Bad Request) response or a 406
(Not Acceptable) response. However I noticed we already had precedence
for 400 for invalid JSON so I kept consistent with that.

Before:
Screenshot 2022-06-28 at 13 11 53

After:
Screenshot 2022-06-28 at 13 12 11

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,12 +1,19 @@
 class ApplicationController < ActionController::API
   include GDS::SSO::ControllerMethods
+  class InvalidRequest < RuntimeError; end
+
   before_action :authenticate_user!
   rescue_from Mongoid::Errors::DocumentNotFound, with: :error_404
+  rescue_from InvalidRequest, with: :error_400
 
 private
 
   def error_404
     head :not_found
+  end
+
+  def error_400
+    head :bad_request
   end
 
   def config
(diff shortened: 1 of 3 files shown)


app/controllers/application_controller.rb
```diff
@@ -47,4 +60,6 @@ def api_url_method
       method(:content_item_url)
     end
   end
+
+  class InvalidRequest < RuntimeError; end
```
brucebolt, line 64: Is this needed here?  It seems to be a duplicate of lines 3 and 11.

kevindew: Oops, thanks for spotting that. Just something I accidentally left over.

brucebolt: It looks like line 11 can be removed too (as the class is already declared on line 3)

brucebolt (approved):

app/controllers/application_controller.rb
kevindew: Argh! sorry. I seem to be an idiot today.

This may have looked like a test of your attention to detail on a PR but it is just me making a mess. Thanks again for catching.

PR #949, opened 2022-07-11
Add rake task to update titles on hmrc_manual_section items.

KludgeKML:
To simplify front-end rendering, we need to add the parent manual title to hmrc manual section items. There's a PR to add this into the schema: 

https://github.com/alphagov/govuk-content-schemas/pull/1106

...and another to do the lookup in the future as the item is written into the content store:

https://github.com/alphagov/hmrc-manuals-api/pull/702

This PR adds a task to update the values on existing records (can be removed once this job is complete and all existing and new hmrc_manual_sections will have the title already).

https://trello.com/c/NToovfjr/1349-add-parent-manuals-title-to-the-hmrc-section-content-item

### DIFF
--- a/lib/tasks/update_hmrc_manual_sections.rake
+++ b/lib/tasks/update_hmrc_manual_sections.rake
@@ -1,0 +1,26 @@
+def update_hmrc_manual_section_titles(dry_run: false)
+  ContentItem.where(document_type: "").each do |ci|
+    manual_content_id = GdsApi.publishing_api.lookup_content_id(ci.details["manual"]["base_path"])
+    if content_id
+      manual_content_id = GdsApi.publishing_api.get_live_content(manual_content_id)
+      ci.details["manual"]["title"] = manual_content_id["title"]
+      puts("Updating HMRC Manual Section #{ci.content_id} to set manual title to #{ci.details["manual"]["title"]}")
+      ci.save unless dry_run
+    else
+      put("ERROR: Couldn't find manual content item for path #{ci.details["manual"]["base_path"]}")
+    end
+  end
+end
+
+namespace :update_hmrc_manual_sections do
+  desc "Show hmrc_manual_section items that will be updated with the title of the containing manual"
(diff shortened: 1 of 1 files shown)


kevindew:
It seems unusual to have a Content Store rake task - since the data is authoritatively held by Publishing API. Can we just use [the downstream tasks](https://github.com/alphagov/publishing-api/blob/49ee90c5bbca13577251415756e989fdef9ed441/lib/tasks/represent_downstream.rake) in Publishing API to deal with this?

KludgeKML:
Possibly? I missed that publishing api was holding the data from hmrc-manuals-api before passing it on, so the downstream tasks would certainly be a better way to update content-store. What's the best way to update the data in publishing-api, though? I see there's mention of data migrations, but also a preference to reduce the number of data migrations. Should I just re-create this task in publishing-api (and use the downstream tasks to then push it to content-store)? Or is there a preferred way?

1pretz1:
Hello 👋 Sorry to also throw may hat in. Agreed that the migration should live in Publishing API (would be HMRC Manuals API but doesn't actually have a DB).

> What's the best way to update the data in publishing-api, though? 

I believe best practice for data migrations in Publishing API is to create a temporary rake task, like the approach in this PR.

As for the implementation, an example of changing an attribute and representing documents downstream can be found in the following PR: https://github.com/alphagov/publishing-api/pull/2047/files which represents all HMRC manual / sections downstream. From looking at the edition in Publishing API, it appears that the `manual` hash (on a section) is in the details JSON. We should just be able to update this hash and re-represent the sections downstream.

```
irb(main):007:0> Edition.where(document_type: "hmrc_manual_section").last.details[:manual]
=> {:base_path=>"/hmrc-internal-manuals/customs-cds-volume-3-tariff-step-by-step-guide"}
```

KludgeKML:
Made obsolete by: https://github.com/alphagov/publishing-api/pull/2112

PR #959, opened 2022-07-28
Remove route-consistency-checker task

callumknights:
Remove route-consistency-checker task along with any associated code

Done as part of CD work, this functionality was meant to pair with an equivalent
task in publishing-api, which has already been removed.

PR removing the publishing-api consistency checker https://github.com/alphagov/publishing-api/pull/2086

Though not immediately obvious, content-dumper is also only used for the above, hence it's removal, outlined here 
https://github.com/alphagov/content-store/pull/276


BeckaL (approved):
Looks good, thanks for adding a bit more detail to the description

PR #961, opened 2022-07-28
Remove export data task

callumknights:
Task no longer needed, was written to import initial data into publishing-api,
those endpoints no longer exist in publishing-api.

Done as part of CD work


BeckaL (approved):
Nice one! Good to clean up some stuff from 2015 😄  Have had a look and can't see any remaining code left hanging around 👍

PR #962, opened 2022-07-28
Remove find_specific_term task

callumknights:
The task doesn't appear to be being used and is untested, and feels sensible to remove as part of the CD work while a bunch of other old rake tasks are being removed. 


JonathanHallam (approved):
thanks for adding the comment :)

PR #968, opened 2022-08-10
New test for the find specific term rake task

callumknights:
This task was added back in after being removed during CD work, didn't realise it was still in use by the content community. So to make content-store compliant with test coverage for CD it needs to be tested adequately.

### DIFF
--- a/app/lib/find_specific_term.rb
+++ b/app/lib/find_specific_term.rb
@@ -3,14 +3,14 @@ require "csv"
 class FindSpecificTerm
   attr_reader :term
 
-  CSV_HEADERS = ["Title", "URL", "Publishing application", "Tagged organisation", "Format", "Content ID"].freeze
+  CONTENT_ITEM_HEADERS = ["Title", "URL", "Publishing application", "Tagged organisation", "Format", "Content ID"].freeze
 
   def initialize(term)
     @term = term
   end
 
   def call
-    write_csv
+    report
   end
 
   def self.call(*args)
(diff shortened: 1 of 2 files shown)


app/lib/find_specific_term.rb
```diff
@@ -19,30 +19,28 @@ def self.call(*args)
...
+      logger.info content_item_fields(content_item).join(", ")
     end
 
+    logger.info "Found #{term_content_items.count} items containing #{term}"
+
     logger.info "Finished searching"
     # There's a subtle bug with RuboCop regexp where it thinks this
     # format of interpolated `Rails.root.join` is incorrect.
-    logger.info "CSV file at #{Rails.root.join('tmp/search_term_content_items.csv')}"
+    # logger.info "CSV file at #{Rails.root.join('tmp/search_term_content_items.csv')}"
```
BeckaL, line 38: Do we still want these lines? I'm imagining so

callumknights: I actually left those comment lines in by mistake, to be honest that entire comment block and the logging is unnecessary now as a CSV is no longer being written. I'll remove it all 👍

brucebolt: Could you also add an explanation into the commit message as to why we're removing the CSV writing as part of this work.

callumknights: Done 👍

BeckaL (approved):
Nice one, good persevering on this. Looks good to me 👍

PR #989, opened 2022-10-05
Revert "Bump plek from 4.0.0 to 4.1.0"

brucebolt:
Reverts alphagov/content-store#987

We are currently filling our log files with deprecation warnings for `Plek.current`, which is causing the machines to run out of disk space:
```
Plek.current is deprecated and will be removed. Use Plek.new or Plek.find instead.
```

Reverting until we replace all uses of this.


JonathanHallam (approved):
👍

PR #1008, opened 2022-11-07
Add task to exclude terms that match document type

leenagupte:
# What's changed?

Adds a task to find content that doesn't match a particular format. 

Usage:

```
bundle exec rake report:find_specific_term_references_excluding_types[Term,"doc_type_1 doc_type_2"]
```

e.g.

```
bundle exec rake report:find_specific_term_references_excluding_types[Licence,"licence aaib_report"]
```

# Why?

Content items aren't always created using the expected format.

### DIFF
--- a/app/lib/find_specific_term.rb
+++ b/app/lib/find_specific_term.rb
@@ -1,13 +1,13 @@
 require "csv"
 
 class FindSpecificTerm
-  attr_reader :term, :exclude_type
+  attr_reader :term, :exclude_types
 
   CONTENT_ITEM_HEADERS = ["Title", "URL", "Publishing application", "Tagged organisation", "Format", "Content ID"].freeze
 
-  def initialize(term, exclude_type = nil)
+  def initialize(term, exclude_types = [])
     @term = term
-    @exclude_type = exclude_type
+    @exclude_types = exclude_types
   end
 
   def call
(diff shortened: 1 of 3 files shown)


app/lib/find_specific_term.rb
```diff
@@ -66,4 +65,12 @@ def content_items(term)
       .or('details.more_info_contact_form': term)
       .or('details.more_info_email_address': term).entries
   end
+
+  def number_of_terms_items
+    term_content_items.count { |content_item| !exclude_types.include?(content_item.document_type) }
```
1pretz1, line 70: Nitpick: we could use `exclude?` instead of the inverse of `include?`

spec/lib/find_specific_term_spec.rb
```diff
@@ -24,5 +26,19 @@
 
       FindSpecificTerm.call(term)
     end
+
+    it "does not find content item if document type is excluded" do
```
1pretz1, line 30: Minor: could we also test that an eligible document will return when excluding certain doc types?

leenagupte:
@1pretz1 Thanks for your review! I've made the recommended changes an fixed up.

1pretz1 (approved):
Nice - looks good to me! 🍏

PR #1010, opened 2022-11-11
Lint: use list syntax for Dockerfile CMD.

sengi:
Generated with `gsed -i 's/^CMD bundle exec puma$/CMD [bundle, exec, puma]/'`


kevindew (approved):
Same comment as per: https://github.com/alphagov/publishing-api/pull/2190#pullrequestreview-1178833765 but not blocking

Issue #1006, opened 2022-11-06
CORS header ‘Access-Control-Allow-Origin’ missing


NickColley:
I have [built a JavaScript API for the Search API and Content APIi](https://github.com/NickColley/govuk), but there is an inconsistency between the two.

In Node.js both succeed as expected.
In the browser only Search API succeeds.
This is because Content API does not have the [correct headers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin) set to allow for client-side requests.

```html
<script type="module">
  import { SearchAPI, ContentAPI } from "https://unpkg.com/@nickcolley/govuk";
  const searchApi = new SearchAPI();
  const searchItem = await searchApi.get("Register to vote")
  // Succeeds
  console.log({ searchItem })
  
  const contentApi = new ContentAPI();
  const contentItem = await contentApi.get("/Register-to-vote")
  // Fails
  console.log({ contentItem })
</script>
```
https://jsbin.com/cidexadexu/edit?html,console,output

Browser console:
```
Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at https://www.gov.uk/api/content/Register-to-vote. (Reason: CORS header ‘Access-Control-Allow-Origin’ missing). Status code: 404.

XHRGEThttps://www.gov.uk/api/content/Register-to-vote
CORS Missing Allow Origin
```

cc @kevindew thoughts on this? :)

kevindew:
I don't think a need has ever come up before, I've opened https://github.com/alphagov/content-store/pull/1011 to add the header as it seems like it should allow this.


I can't make any guarantees for future consistency on this, I don't think anyone is taking our API offerings seriously as external resources.

aldavidson:
Can we close this issue since #1011 was merged?

NickColley:
Yes it all is working!

PR #1011, opened 2022-11-14
Allow cross-origin requests to access public content endpoint

kevindew:
This addresses https://github.com/alphagov/content-store/issues/1006

This sets a `Access-Control-Allow-Origin: *` header on HTTP responses to the show endpoint, this will allow requests to this API to originate (via JS) from hosts other than www.gov.uk.

The reason for adding this is resolve the issue raised on GitHub 1006 [1]. GOV.UK doesn't directly need this as it doesn't make use of client side requests to the Content Store and, if it did, they'd be from the same host. However this is added in to reflect that this is indeed a partially supported public API [2] and that we are not concerned with JS clients calling it from a different host.

I only put this on the one endpoint as there didn't seem to be any need to have it on endpoints other than ContentItems#show. I also didn't implement the HTTP OPTIONS method for the endpoint as I don't think it's strictly needed and can't see evidence of this enabled on other GOV.UK cross-origin endpoints:

➜  ~ curl -sI -X OPTIONS https://www.gov.uk/api/search.json\?fields\=publishing_app\&filter_publishing_app\=publisher\&filter_first_published_at\=from:2022-09-01,to:2022-09-30 | grep HTTP HTTP/2 404

[1]: https://github.com/alphagov/content-store/issues/1006
[2]: https://content-api.publishing.service.gov.uk/

### DIFF
--- a/app/controllers/content_items_controller.rb
+++ b/app/controllers/content_items_controller.rb
@@ -1,6 +1,7 @@
 class ContentItemsController < ApplicationController
   skip_before_action :authenticate_user!, only: [:show]
   before_action :parse_json_request, only: [:update]
+  before_action :set_cors_headers, only: [:show]
 
   def show
     item = GovukStatsd.time("show.find_content_item") do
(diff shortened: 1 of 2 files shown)


BeckaL (approved):
LGTM 👍

kevindew:
Great, thanks for reporting back. I'd totally forgotten about this so glad to have confirmation it works.

PR #1015, opened 2022-11-16
Add a feature flag to disable modifying Router backends on the fly.

sengi:
Content Store currently updates the backend address in Router every time a content item or publishing intent is saved. This was originally done [to simplify a long-since-completed migration](https://github.com/alphagov/content-store/pull/23), but it's now both unnecessary and causes problems with parallel running for Replatforming, where the backend addresses necessarily differ between the old and new installations.

This lets us disable this esoteric behaviour on the Kubernetes side during the Replatforming pilot phase.

Tested: verified that router_api.add_backend is called when the NO_SET_ROUTER_BACKEND_ADDRESS_ON_PUBLISH env var is not set and not otherwise.

### DIFF
--- a/app/models/route_set.rb
+++ b/app/models/route_set.rb
@@ -67,7 +67,14 @@ class RouteSet < OpenStruct
         register_gone_route(route)
       end
     else
-      register_rendering_app
+      # TODO: clean up NO_SET_ROUTER_BACKEND_ADDRESS_ON_PUBLISH after
+      # Replatforming MVP launch (i.e. once GOV.UK is running on Kubernetes in
+      # production). This could be done by removing the flag (i.e. kicking the
+      # can down the road) or by removing the concept of dynamically
+      # configuring Router backends entirely (probably better), or perhaps even
+      # something else.
+      truthy_env_values = /^[1ty]/i
+      register_rendering_app unless truthy_env_values.match?(ENV["NO_SET_ROUTER_BACKEND_ADDRESS_ON_PUBLISH"])
 
       routes.each do |route|
         register_route(route, rendering_app)
--- a/spec/models/route_set_spec.rb
+++ b/spec/models/route_set_spec.rb
@@ -170,6 +170,26 @@ describe RouteSet, type: :model do
       route_set.register!
     end
 
+    describe "when NO_SET_ROUTER_BACKEND_ADDRESS_ON_PUBLISH=1" do
+      around do |t|
+        ClimateControl.modify NO_SET_ROUTER_BACKEND_ADDRESS_ON_PUBLISH: "1" do
+          t.run
+        end
+      end
+
+      it "does not call router_api.add_backend" do
+        expect(Rails.application.router_api).not_to receive(:add_backend)
+        expect(Rails.application.router_api).to receive(:commit_routes)
+
+        route_set = RouteSet.new(base_path: "/path", rendering_app: "frontend")
+        route_set.routes = [
+          { path: "/path", type: "exact" },
+          { path: "/path/sub/path", type: "prefix" },
+        ]
+        route_set.register!
+      end
+    end
+
     it "registers and commits all registerable redirects for a redirect item" do
       redirects = [
         { path: "/path", type: "exact", destination: "/new-path" },


config/initializers/feature_flags.rb
```diff
@@ -0,0 +1,5 @@
+# Every time a content item is published, content-store updates Router's
+# backend address for that content item's rendering app.
+# If the environment variable NO_SET_ROUTER_BACKEND_ADDRESS_ON_PUBLISH is set,
+# content-store will not update backend addresses on publish.
+SET_ROUTER_BACKEND_ADDRESS_ON_PUBLISH = !ENV["NO_SET_ROUTER_BACKEND_ADDRESS_ON_PUBLISH"]
```
kevindew, line 5: Since we don't have a common concept of `config/initializers/feature_flags` I'd imagine it's better to have this just directly in app/models/route_set.rb and avoid the need to search the codebase for where the constant is set and to learn more.

I'd also include in the comment that this is added temporarily for the 2020 replatforming project and is expected to not be needed once that is complete.

richardTowers, line 5: There's a possible future source of confusion here where someone does:

```diff
- NO_SET_ROUTER_BACKEND_ADDRESS_ON_PUBLISH=true
+ NO_SET_ROUTER_BACKEND_ADDRESS_ON_PUBLISH=false
```

but `!"false"` is `false` in ruby.

Suggest:

```suggestion
SET_ROUTER_BACKEND_ADDRESS_ON_PUBLISH = ENV["NO_SET_ROUTER_BACKEND_ADDRESS_ON_PUBLISH"] != "true"
```

app/models/route_set.rb
```diff
@@ -67,7 +67,9 @@ def register!
         register_gone_route(route)
       end
     else
-      register_rendering_app
+      if SET_ROUTER_BACKEND_ADDRESS_ON_PUBLISH
+        register_rendering_app
+      end
```
richardTowers, line 72: Arguable small stylistic improvement:

```suggestion
      register_rendering_app if SET_ROUTER_BACKEND_ADDRESS_ON_PUBLISH
```

you could also consider skipping the negation and going with

```suggestion
      register_rendering_app unless NO_SET_ROUTER_BACKEND_ADDRESS_ON_PUBLISH
```

which I think reads okay, but is probably personal taste

richardTowers (commented):
Couple of drive-by nitpicks (sorry)

app/models/route_set.rb
kevindew, line 72: This looks good. It'd be good to have a test with it though

I imagine something like: 

```
    it "doesn't register a backend when NO_SET_ROUTER_BACKEND_ADDRESS_ON_PUBLISH environment variable is set" do
        ClimateControl.modify NO_SET_ROUTER_BACKEND_ADDRESS_ON_PUBLISH: "true" do
            backend_request = stub_request(:put, %r{\A#{GdsApi::TestHelpers::Router::ROUTER_API_ENDPOINT}/backends/})
            @item.register_routes
            expect(backend_request).not_to have_been_requested
        end
    end
```

kevindew (commented):
This looks good, just a couple of small changes.

config/initializers/feature_flags.rb
sengi: Sure, no probs. I was going for "as simple and predictable as possible" but happy to go with your suggestion. (It still wouldn't handle "no" or "0", which are also quite commonly used and could reasonably be expected to work, but since it's probably only going to be used the once and only by Replatforming, I think that's fine. FWIW I'd usually go with `/^[1ty]/i.match?` but the prevailing style in this repo seemed to be `if ENV['foo']` hence I stuck with that style.)

app/models/route_set.rb
sengi: @kevindew I'm not quite seeing how the proposed test would invoke the initializer. (That's partly why I didn't commit the bodgy one-off test that I wrote, though also I considered that it wasn't worth the overhead of maintaining a test just for a feature flag)

config/initializers/feature_flags.rb
sengi: Ah ok, fair enough. I'll do that.

sengi: so some of the other boolean env vars in this codebase already have that potential issue, e.g. `MONGO_SSL=false` enables TLS on the `mongoid` client.

sengi: oh wait no that's not a valid example, didn't notice it was ERB

sengi: RAILS_LOG_TO_STDOUT and RAILS_SERVE_STATIC_FILES are example of that though.

richardTowers: Yeah, your regex looks good / better - happy if you want to go for that.

I'm strongly of the view that testing for the presence of an env var is dangerous in general, because of the `VARIABLE=false` being evaluated to "yep, VARIABLE is non-empty, let's goooo". Anything that checks the actual value to some degree is a big improvement on that - `= "true"` is the simplest (but most naive) thing.

I wonder if I should put something in the style guide about this 🤔

app/models/route_set.rb
sengi: (just finished reading the rest of the comments - that test makes sense if we're reading the env var in the method)

config/initializers/feature_flags.rb
sengi: I agree — I think these things should follow the principle of least surprise! I'll go with the regex then 👍

sengi:
RFAL.

richardTowers (approved):

richardTowers:
> Rails default config goes for just using presence so it'll always be hard to fight that precedence.

Oof, today I learned. 😞

sengi:
Thanks both!

PR #1046, opened 2023-01-30
Allow verify-pact workflow to accept an artifact input

kevindew:
This provides an alternative method for this workflow to execute
Publishing API pact tests. Rather than accessing the Pact tests via the
internet, downloading them from the pact-broker, this instead allows an
input of an artifact filename. This action will download that artifact
and run against the pacts.

The motivation for this approach is that it can allow the Publishing API
CI [1] to execute these tests without the initiator of a Publishing API
needing access to write to the GOV.UK Pact Broker. This is a problem we
are currently experiencing with Dependabot as it does not have access to
GitHub Action secrets.

At the moment, and likely for the forseeable future, there isn't a
scenario where we'd need to use any file other than
`publishing_api-content_store.json` so I've hardcoded this. We can
always change it to a parameter in future if there is a need.

[1]: https://github.com/alphagov/publishing-api/blob/a4bf641a571b8eeeb61f3ca1640a1ff26233423e/.github/workflows/ci.yml#L97-L103

---

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,4 +27,8 @@ jobs:
     name: Run Pact tests
     uses: ./.github/workflows/verify-pact.yml
     with:
+      # Note: As of January 2023 this relies on Jenkins CI to generate
+      # this Pact consumer version. Replatforming does not tag
+      # deployed-to-production and the Publishing API CI build doesn't run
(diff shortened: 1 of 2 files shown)


kevindew:
Thanks @brucebolt, we should probably do something at tech fortnightly to help share knowledge of actions more.

PR #1054, opened 2023-03-15
removing schedule as this is now captured in EKS environment

martynipratt:
https://github.com/alphagov/govuk-helm-charts/blob/main/charts/app-config/values-integration.yaml#L646

State captured in the EKS environment removing to prevent clash.


martynipratt:
meant this to be a draft. will just reopen when it is required

PR #1074, opened 2023-05-19
Mock out router-api calls when DISABLE_ROUTER_API is true

aldavidson:
Allows [router-api](https://github.com/alphagov/router-api) calls to be disabled by setting this environment variable. 
This then allows us to avoid the potential for conflicts when we deploy both the Mongo and [PostgreSQL](https://github.com/alphagov/content-store-on-postgresql) versions simultaneously behind a [mirroring proxy](https://github.com/alphagov/content-store-proxy), during the migration of content-store from Mongo to PostgreSQL. In this scenario, only the primary will update routes in Router API, and the secondary will not.

[Trello card](https://trello.com/c/pgBHpAZ7/616-put-all-router-api-updates-behind-a-feature-flag-in-content-store) / [Overall Epic](https://trello.com/c/C1BQDFTG/502-plan-for-migrating-content-store-off-mongodb)

### DIFF
--- a/app/lib/mock_router_api.rb
+++ b/app/lib/mock_router_api.rb
@@ -3,33 +3,19 @@
 # This is to allow dual-running of both MongoDB and PostgreSQL Content Stores
 # for a transition period during the migration to PostgreSQL
 class MockRouterApi
-  def delete_route(*args)
-    log("delete_route", args)
+  # Whatever the application tries to send to router-api,
+  # just log it and do nothing else
+  def method_missing(method_name, *args, **kwargs, &_block)
+    log(method_name.to_s, args, kwargs)
   end
 
-  def add_backend(*args)
-    log "add_backend", args
-  end
-
-  def add_redirect_route(*args)
-    log "add_redirect_route", args
-  end
-
-  def add_gone_route(*args)
-    log "add_gone_route", args
-  end
-
-  def add_route(*args)
-    log "add_route", args
-  end
-
-  def commit_routes(*args)
-    log "commit_routes", args
+  def respond_to_missing?(method, *)
+    GdsApi.router.methods.include?(method) || super
   end
 
 private
 
-  def log(method, args)
-    logger.debug "Mocked call to router_api: #{method}(#{args})"
+  def log(method, *args, **kwargs)
+    logger.debug "Mocked call to router_api: #{method}(#{args}, #{kwargs})"
   end
 end
--- a/app/lib/mock_router_api.rb
+++ b/app/lib/mock_router_api.rb
@@ -1,0 +1,35 @@
+# When DISABLE_ROUTER_API is set to true, the application will use this
+# class instead of GdsApi::Router
+# This is to allow dual-running of both MongoDB and PostgreSQL Content Stores
+# for a transition period during the migration to PostgreSQL
+class MockRouterApi
+  def delete_route(*args)
+    log("delete_route", args)
+  end
+
+  def add_backend(*args)
+    log "add_backend", args
+  end
+
+  def add_redirect_route(*args)
+    log "add_redirect_route", args
+  end
+
+  def add_gone_route(*args)
+    log "add_gone_route", args
+  end
+
+  def add_route(*args)
+    log "add_route", args
+  end
+
+  def commit_routes(*args)
+    log "commit_routes", args
+  end
+
+private
+
+  def log(method, args)
+    logger.debug "Mocked call to router_api: #{method}(#{args})"
+  end
+end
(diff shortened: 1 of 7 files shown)


config/application.rb
```diff
@@ -113,10 +113,18 @@ class Application < Rails::Application
...
+    end
+
+    def new_router_api_adapter
+      if ENV["DISABLE_ROUTER_API"].to_s == "true"
+        MockRouterApi.new
+      else
+        GdsApi::Router.new(
+          Plek.find("router-api"),
+          bearer_token: ENV["ROUTER_API_BEARER_TOKEN"] || "example",
+        )
```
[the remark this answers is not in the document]
kevindew: If you were to go further (which I don't think is necessary) - this app uses a rather legacy pattern of setting up the adapater for the lifespan of the application. Our more modern approach has been to just create adapters on demand when they're needed.

So if it happened to be simpler for you, you could remove the code from here and then just pick the adapter in the class that talks to Router API. However definitely not necessary for this PR just if it helps.

config/application.rb
```diff
@@ -113,10 +113,18 @@ class Application < Rails::Application
...
     def router_api
-      @router_api ||= GdsApi::Router.new(
-        Plek.find("router-api"),
-        bearer_token: ENV["ROUTER_API_BEARER_TOKEN"] || "example",
-      )
+      @router_api ||= new_router_api_adapter
+    end
+
+    def new_router_api_adapter
+      if ENV["DISABLE_ROUTER_API"].to_s == "true"
```
kevindew, line 120: I think the `.to_s` is redundant here.

app/lib/mock_router_api.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  def commit_routes(*args)
+    log "commit_routes", args
+  end
+
+private
+
+  def log(method, args)
+    logger.debug "Mocked call to router_api: #{method}(#{args})"
```
kevindew, line 33: Should we be explicit and use Rails.logger - I hadn't actually realised that logger was a global method (does ActiveSupport patch it into Object?)

app/lib/mock_router_api.rb
```diff
@@ -0,0 +1,35 @@
+# When DISABLE_ROUTER_API is set to true, the application will use this
+# class instead of GdsApi::Router
+# This is to allow dual-running of both MongoDB and PostgreSQL Content Stores
+# for a transition period during the migration to PostgreSQL
+class MockRouterApi
```
kevindew, line 5: I wonder if you might want to simplify this class to be just have `def method_missing`  and log that, then you'd not need to be considering the actual API of the adapter and could have much fewer tests

app/lib/mock_router_api.rb
kevindew: Also, just wanted to sense choice of debug level, since that'll only run in dev and test environments, not production. Is that correct that you wouldn't want this to log when run in any of the GOV.UK environments? (I don't have an opinion on which you choose)

kevindew (commented):
Nice, I've added a few comments to question a couple of things and offer a few simplification routes.

config/application.rb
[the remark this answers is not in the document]
aldavidson: Yeah, I thought it was a bit of an awkward way of doing it, but was trying to follow a 'smallest change necessary' approach. If you're OK with this PR doing a bit of refactoring-as-we-go, I'll gladly do that.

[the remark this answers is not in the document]
aldavidson: I've pushed #ac4079b5 to cover this and the comment above

app/lib/mock_router_api.rb
aldavidson: Yeah, that's a fair point - `info` might be better while we're testing on integration and staging. We can take a judgement call before rolling out to production as to whether we want to adjust it back down again then.

app/lib/mock_router_api.rb
aldavidson: done

app/lib/mock_router_api.rb
aldavidson: done

app/lib/mock_router_api.rb
```diff
@@ -0,0 +1,21 @@
...
+  # Whatever the application tries to send to router-api,
+  # just log it and do nothing else
+  def method_missing(method_name, *args, **kwargs, &_block)
+    log(method_name.to_s, args, kwargs)
+  end
+
+  def respond_to_missing?(method, *)
+    GdsApi.router.methods.include?(method) || super
```
kevindew, line 13: Nice 👍

PR #1088, opened 2023-06-14
Add rake task to export named/all mongodb collections to JSON files

aldavidson:
Part of setting up [an overnight export/import between Mongo and PostgreSQL](https://trello.com/c/GMoOthyj/671-add-step-to-overnight-environment-sync-job-to-export-mongodb-to-json-and-import-to-postgresql), as part of the overall epic [Migrate content-store to PostgreSQL](https://trello.com/c/C1BQDFTG/502-plan-for-migrating-content-store-off-mongodb)


aldavidson:
Have requested @sengi for review because of the direct way it has to install the mongo package (i.e. not from a standard ubuntu package repo)

Dockerfile
```diff
@@ -14,6 +14,11 @@ RUN bootsnap precompile --gemfile .
 
 FROM $base_image
 
+# Install `mongodb-database-tools` for 'mongoexport'
+RUN wget -P /tmp/ https://fastdl.mongodb.org/tools/db/mongodb-database-tools-ubuntu1604-x86_64-100.7.2.deb && \
+    apt-get install -y /tmp/mongodb-database-tools-ubuntu1604-x86_64-100.7.2.deb && \
+    rm /tmp/mongodb-database-tools-ubuntu1604-x86_64-100.7.2.deb
```
sengi, line 20:
```suggestion
# TODO: remove this temporary MongoDB package once we no longer need mongoexport (once the migration to Postgres is done).
ARG mongo_package=mongodb-database-tools-ubuntu2204-x86_64-100.7.2.tgz
ARG mongo_package_repo=https://fastdl.mongodb.org/tools/db
WORKDIR /tmp
ENV DEBIAN_FRONTEND=noninteractive
RUN wget --progress=mega "${mongo_package}/${mongo_package_repo}" && \
    apt-get install -y --no-install-recommends "${mongo_package}" && \
    rm -fr /tmp/*
```

sengi: Normally you'd add the apt repo so that future builds will pick up security patches and whatnot, but since this is just a temporary thing to add a tool for the migration and updating it isn't really an issue, I think it's a good call to just install the package directly 👍 (This way we don't slow down the build anything like as much, so hopefully happier developers.)

The suggestion has a bunch of changes for readability/maintainability/robustness:

- Use the package that's built for the base image's Ubuntu version (2204 not 1604).
- Use WORKDIR to change directory in the target filesystem.
- Use non-interactive settings for command-line tools (and use the wget progress setting that doesn't blow out the build logs).
- DRY up pathnames, URLs etc.

lib/tasks/export.rake
```diff
@@ -0,0 +1,13 @@
+namespace :mongo do
```
sengi, line 1: Might wanna stick a TODO on this as well (assuming it's meant to be cleaned up after the migration)

sengi (approved):

Dockerfile
sengi: Oh I'm assuming this doesn't need to work on arm64 cos it's only ever gonna run in prod, right? (and only for a couple of months so no chance of it running into any move to ARM there)

lib/tasks/export.rake
aldavidson: Yeah, when it's clean-up time we'll basically do a global find for 'mongo' and delete pretty much everything. Happy to add a TODO though

chrisroos (approved):
Looks good to me, @aldavidson :+1:

PR #1096, opened 2023-06-24
Present content-item JSON with deep-sorted keys

aldavidson:
... for equivalence with `port-to-postgres` version

When dual-running content-store on MongoDB (main) and PostgreSQL (port-to-postgresql branch), we find that [the JSON generated by the two versions differs](https://trello.com/c/1gljPBwl/720-find-out-why-the-contentitem-links-field-differs-between-mongo-pg). After much investigation, it seems that the remaining source of difference is the order of keys in the content_items - particularly in the `links` JSON.

The source of this difference is that Mongo stores the data as BSON, which [preserves the order of the JSON it is given](https://softwareengineering.stackexchange.com/a/340269](https://softwareengineering.stackexchange.com/a/340269) ):

> A BSON data structure is an ordered object, not a dictionary  

whereas in PostgreSQL we store the JSON as `JSONB`, which [sorts the keys of an object on write](https://www.postgresql.org/docs/current/datatype-json.html](https://www.postgresql.org/docs/current/datatype-json.html) ):

> jsonb does not preserve white space, does not preserve the order of object keys, and does not keep duplicate object keys

As a result, even when the JSON is functionally equivalent, a character-by-character comparison of the two will fail. To have a good level of confidence in the [Mongo-to-PostgreSQL migration](https://trello.com/c/C1BQDFTG/502-plan-for-migrating-content-store-off-mongodb), we need to be able to demonstrate that the JSON from the two versions is exactly equivalent.

So this PR introduces the `deepsort` gem, and a trivial wrapper class to make sure that we don't accidentally sort nested array elements as well. We use it to sort content-items and their expanded_links.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,6 +3,7 @@ source "https://rubygems.org"
 gem "rails", "7.0.5"
 
 gem "bootsnap", require: false
+gem "deepsort"
 gem "gds-api-adapters"
 gem "gds-sso"
 gem "govuk_app_config"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -103,6 +103,7 @@ GEM
       database_cleaner-core (~> 2.0.0)
       mongoid
     date (3.3.3)
+    deepsort (0.4.5)
     diff-lcs (1.5.0)
     docile (1.4.0)
     domain_name (0.5.20190701)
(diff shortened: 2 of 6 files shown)


JonathanHallam (approved):
Looking good, thanks Al!

PR #1107, opened 2023-07-04
Fix syntax error in deploy.yml workflow.

sengi:
I didn't spot this in #1086. GitHub really ought to syntax-check these automatically pre-merge.


aldavidson (approved):
Good catch, thanks!

PR #1116, opened 2023-07-19
Provide fallback value for ecrRepositoryName

CristinaRO:
Commit 5c9e9f6af13a18e17e8cf7d7cdd8bd3d6e5bb2c1 added a choice of ECR repository name on build-and-publish-image.

One might think that the default value of the input is there to provide a fallback for non-workflow_dispatch triggered jobs.

One might, apparently, be wrong[1].

Automatic deploys on content-store have been failing since the choice was introduced, while manual deploys are succeeding. The error is that the image tag is invalid, and it is invalid because it is missing the value of `ECR_REPOSITORY`.

Thanks to Bruce for identifying the related commits, which provided me with a clue of where to look for answers!

[1] https://dev.to/mrmike/github-action-handling-input-default-value-5f2g

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -39,7 +39,7 @@ jobs:
     uses: alphagov/govuk-infrastructure/.github/workflows/build-and-push-image.yml@main
     with:
       gitRef: ${{ inputs.gitRef || github.ref }}
-      ecrRepositoryName: ${{ inputs.ecrRepositoryName }}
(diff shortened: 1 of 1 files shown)


theseanything (approved):
Good catch! Yup "inputs.ecrRepositoryName" will only be populated on `workflow_dispatch` triggered runs.

sengi (approved):
Looks reasonable to me. @theseanything might be able to spot any gotchas I might have missed :)

PR #1138, opened 2023-08-29
Gzip Mongo exports

aldavidson:
In staging, when run as part of the Kubernetes CronJob [content-store-mongo-to-postgres-cron](https://github.com/alphagov/govuk-helm-charts/blob/main/charts/govuk-jobs/templates/content-store-mongo-to-postgres-cronjob.yaml) the job fails with this error -
`'The node was low on resource: ephemeral-storage. '`

- i.e. the node it was running on, ran out of storage space (see [discussion on slack](https://gds.slack.com/archives/C013F737737/p1693301007257509))

This PR pipes all files produced by the  `mongo:export` rake tasks through gzip, which produces pretty good compression on the largest collections (content_items particularly) and should drastically cut down the amount of ephemeral storage it needs.

[Trello card](https://trello.com/c/U4kiudWW/820-fix) , see also #1139

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,9 @@
 # Ignore the default SQLite database.
 /db/*.sqlite3
 
+# Ignore mongo DB dumps
+/db/mongo-dumps
+
 # Ignore all logfiles and tempfiles.
 /log/*.log
 /tmp
--- a/app/lib/mongo_exporter.rb
+++ b/app/lib/mongo_exporter.rb
@@ -5,12 +5,13 @@ class MongoExporter
 
   def self.export(collection:, path:)
     FileUtils.mkdir_p(path)
+    zipped_file_path = File.join(path, [collection, "json", "gz"].join("."))
     execute(
       "mongoexport",
       "--uri=#{ENV['MONGODB_URI']}",
       "--collection=#{collection}",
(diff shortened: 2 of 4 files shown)


app/lib/mongo_exporter.rb
```diff
@@ -9,8 +9,8 @@ def self.export(collection:, path:)
       "mongoexport",
       "--uri=#{ENV['MONGODB_URI']}",
       "--collection=#{collection}",
-      "--out=#{File.join(path, [collection, 'json'].join('.'))}",
       "--type=json",
+      " | gzip > #{File.join(path, [collection, 'json', 'gz'].join('.'))}",
```
sengi, line 13: optional: consider assigning the filename to a variable, to take the filename-constructing logic out of the string and improve readability.

.gitignore
```diff
@@ -10,6 +10,9 @@
 # Ignore the default SQLite database.
 /db/*.sqlite3
 
+# Ignore mongo DB dumps
+/db/mongo-dumps
```
sengi, line 14: consider adding this to .dockerignore too

sengi (approved):

app/lib/mongo_exporter.rb
aldavidson: done

PR #1145, opened 2023-08-31
FIx silent failure in mongoexport call

aldavidson:
Since adding a pipe through gzip to the mongo export task in #1138, although the CronJob has been completing without errors, it has not actually been exporting anything. Investigation reveals that the way we're executing the command (with Kernel.system) doesn't like the pipe, and this results in the `mongoexport` command failing, but the error is caught by the `system` call and not bubbled out (it just returns a `nil` value).

[Trello card](https://trello.com/c/sxWnwtoO/822-fix-silent-failure-in-content-store-mongo-export)

This PR changes the way we execute the command to use backticks around a plain string, which 

a) works well in local tests under govuk-docker, and
b) raises an exception if the command returns an error status

### DIFF
--- a/app/lib/mongo_exporter.rb
+++ b/app/lib/mongo_exporter.rb
@@ -1,3 +1,5 @@
+require "open3"
+
 class MongoExporter
   def self.collection_names
     Mongoid.default_client.collections.map(&:name).reject { |e| e == "data_migrations" }.sort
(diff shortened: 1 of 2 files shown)


sengi (approved):

app/lib/mongo_exporter.rb
```diff
@@ -21,7 +21,7 @@ def self.export_all(path:)
     end
   end
 
-  def self.execute(cmd, *args)
-    system cmd, *args
+  def self.execute(*args)
+    `#{args.join(" ")}`
```
sengi, line 25: Ah brakeman doesn't like this, which I guess makes total sense. I'd probably just drop this function and do the .join inline — that way at least if you have to suppress the warning it'll be easier to see that the code is ok and there's no chance of some poor future soul calling this function with untrusted inputs.

aldavidson: Yeah, I found a Brakeman-friendly way of piping commands together with `Open3.pipeline(cmd1, cmd2, ...)` and changed it to that in [3ae3c99](https://github.com/alphagov/content-store/pull/1145/commits/3ae3c99abf4c952f98f950d10b7cb392e4378811)

PR #1152, opened 2023-09-13
Remove placeholder logic

brucebolt:
Placeholder schemas were added to hold documents that were rendered by Whitehall and not yet ready to be presented to end users.

This added some complexity to Publishing API and Content Store, as these routes did not need to be registered with Router.

Therefore removing this logic now all rendering has been migrated out of Whitehall.

Depends on https://github.com/alphagov/publishing-api/pull/2504.

[Trello card](https://trello.com/c/xAZ7nU8h)


app/models/content_item.rb
```diff
@@ -219,12 +219,10 @@ def route_set
 private
 
   def should_register_routes?(previous_item: nil)
-    return false if schema_name.start_with?("placeholder")
-
     if previous_item
```
jkempster34, line 222: Could possibly inline this, but that may be personal preference

jkempster34 (approved):
Looks great

app/models/content_item.rb
brucebolt: Nice idea.  Have made that change.

PR #1173, opened 2023-11-13
Upgrade Rails to 7.1.2 (for Postgres branch)

brucebolt:
This bumps rails to 7.1.2, for the Postgres branch of the application only.

The line specifying `config.active_support.cache_format_version = 7.0` can be removed when we are certain that Rails 7.1 is running correctly in production.  Setting this config allows us to revert back to Rails 7.0.x, should we need to.

Depends on https://github.com/alphagov/content-store/pull/1171.

[Trello card](https://trello.com/c/DVOe4m28)


aldavidson (approved):
LGTM :+1: Thanks!

PR #1190, opened 2023-12-05
Don't force incoming HTTP requests to be over SSL

mike29736:
I assume this was accidentally changed during the Rails upgrade (the
latest version bump seemed to want to touch quite a lot of config).

The traffic to the actual Rails app in non-development environments is
internal to our cluster, so it won't be over SSL.


mike29736:
I've deployed this to integration and it booted and passed the smoke tests

PR #1227, opened 2024-01-31
Remove `simplecov-rcov` gem dependency

brucebolt:
This gem is not being initialised in the `test_helper.rb` file, so therefore is not being used and can be removed.


mec (approved):
🚢

PR #1248, opened 2024-03-12
Remove all content items rendered by Whitehall Frontend

brucebolt:
Whitehall Frontend was switched off in 2023. However, as a result of [a bug](https://trello.com/c/z5pHDAXC) where some documents were missed when they were bulk republished, a small number (approximately 65) of documents remained in the Content Store database with a rendering app of `whitehall-frontend` and a schema name of `placeholder`.

The logic to deal with placeholders was removed in https://github.com/alphagov/content-store/pull/1152, therefore the routes were never registered for these content items (and never will be, as Content Store assumes they have already been rendered).

Since the public versions of these page already return a 404 or 500 error (or in some cases, have a redirect implemented directly in Router), it is safe to remove them and republish the affected pages from Whitehall to correctly register the routes and make these pages available again to users.

[Trello card](https://trello.com/c/UdAr5zOb)


jkempster34 (approved):
LGTM

PR #1357, opened 2024-12-03
Remove route registration to Router API

theseanything:
Router API is no longer used by Router to populate the routing data structure. Instead this is loaded directly from Content Store's database.


app/models/content_item.rb
```diff
@@ -119,21 +113,6 @@ def gone?
     schema_name == "gone" && details_is_empty?
   end
 
-  def router_rendering_app
-    # This is an extension of the hack in `gone?` method.
-    #
-    # For items that are registered in the content store which are not redirects
-    # or gones we need to have a rendering_app. This is fine for all but the
-    # "gone but not gone" exception defined in `gone?` where rendering_app is
-    # not part of the schema for a gone but is required to register the route.
```
sihugh, line 128: 😬

theseanything: Yeah this is for our gross "Gone" pages that show an explanation (they are 200s). This logic has been moved over to router.

https://github.com/alphagov/router/blob/3831c58d1999a1bfc80db6725c091ba0548b92da/lib/routes.go#L26
https://github.com/alphagov/router/blob/3831c58d1999a1bfc80db6725c091ba0548b92da/lib/routes.go#L49

PR #1361, opened 2024-12-09
Notify Router only on specific column updates

theseanything:
Currently, we send a notification to the Router to update route tables whenever there is any change to the content_items (or publish_intents) tables. However, the Router only needs to update the route table when changes affect specific columns that determine the routes.

This change should reduce unnecessary reloading.

### DIFF
--- a/db/migrate/20241209132444_update_notify_trigger_for_route_changes.rb
+++ b/db/migrate/20241209132444_update_notify_trigger_for_route_changes.rb
@@ -1,0 +1,47 @@
+class UpdateNotifyTriggerForRouteChanges < ActiveRecord::Migration[7.2]
+  def up
+    execute <<-SQL
+      CREATE OR REPLACE FUNCTION notify_route_change() RETURNS trigger AS $$
+      BEGIN
+          -- Trigger on INSERT or DELETE
+          IF (TG_OP = 'INSERT' OR TG_OP = 'DELETE') THEN
+              PERFORM pg_notify('route_changes', '');
+              RETURN COALESCE(NEW, OLD);
+          END IF;
+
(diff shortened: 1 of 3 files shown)


db/migrate/20241209132444_update_notify_trigger_for_route_changes.rb
```diff
@@ -0,0 +1,47 @@
...
+          IF (TG_OP = 'INSERT' OR TG_OP = 'DELETE') THEN
+              PERFORM pg_notify('route_changes', '');
+              RETURN COALESCE(NEW, OLD);
+          END IF;
+
+          -- Trigger on UPDATE for specific columns
+          IF (TG_OP = 'UPDATE') THEN
+              IF TG_TABLE_NAME = 'content_items' THEN
+                -- Specific column checks for the content_items table
+                IF (NEW.routes IS DISTINCT FROM OLD.routes OR
```
sihugh, line 16: Is there a scenario where we'd want to push an update to fix a problem in the cached routes? How would we do that?

samsimpson1 (approved):

db/migrate/20241209132444_update_notify_trigger_for_route_changes.rb
theseanything: If I understand correctly, you're asking if there was a problem with a route in the Router's in-memory triemux, but the content store was correct, how do we could we update Router?

We can force a reload via [Router's API `/reload`.](https://github.com/alphagov/router/blob/3831c58d1999a1bfc80db6725c091ba0548b92da/lib/router_api.go#L14) Also routes are periodically reloaded every 5 mins anyways.

PR #1374, opened 2025-01-03
Fix OpenAPI 3.0 definition

brucebolt:
The OpenAPI 3.0 definition was invalid. This can be confirmed by using tools such as the [Swagger Editor](https://editor.swagger.io/) or by looking at an example of where this data was consumed (e.g. [GOV.UK Content API docs](https://content-api.publishing.service.gov.uk/reference.html#contentitem)).

Screenshot of the definition being consumed by the GOV.UK Content API docs.  The example contains the code that refers to the example block, rather than the example itself.

Updating to make this definition valid and adding a test so we don't accidentally make the definition invalid in the future.

[Trello card](https://trello.com/c/7LqhcraV)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -19,6 +19,7 @@ group :development, :test do
   gem "factory_bot"
   gem "govuk_schemas"
   gem "govuk_test"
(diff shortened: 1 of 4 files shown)


JonathanHallam (approved):
All good, thanks

PR #1416, opened 2025-04-01
Turn off production-like JSON logging in dev env

mike3985:
All of our other apps seem to use JSON logging in deployed environments but not in development and this change updates Content Store to follow that same pattern.

Configuring GovukJsonLogging without also setting the GOVUK_RAILS_JSON_LOGGING environment variable doesn't properly enable JSON logging. One result of this is that no Rails request logs are ever printed to stdout or to the local file in log/.


config/initializers/logstasher.rb
```diff
@@ -1,5 +1,7 @@
-GovukJsonLogging.configure do
-  add_custom_fields do |fields|
-    fields[:govuk_dependency_resolution_source_content_id] = request.headers["GOVUK-Dependency-Resolution-Source-Content-Id"]
+if !Rails.env.development? || ENV["GOVUK_RAILS_JSON_LOGGING"].present?
```
yndajas, line 1: **suggestion:** would `unless Rails.env.development? && ENV["GOVUK_RAILS_JSON_LOGGING"].blank?` do the same here? If (not something) or (something) hurts my brain a little

yndajas (approved):

config/initializers/logstasher.rb
mike3985: Ditto... "unless X and Y" is my worst nightmare 😆 Any alternative alternatives?

mike3985: We only want this configuration:

1. in non-development environments, and
2. in the development environment when that variable has been intentionally set

mike3985: Though, come to think of it, that variable is always set in non-development environments... It seems safe to remove the Rails env check

yndajas: Don't we still want to check it in dev though?

mike3985: Yep. I meant that we can replace that conditional with just
```ruby
if ENV["GOVUK_RAILS_JSON_LOGGING"].present?
```

yndajas: Oh 🙈

PR #1589, opened 2026-04-07
[PP-7422] Upgrade to Rails 8.1

yndajas:
This is a fairly straightforward upgrade with decisions documented in commits. Notably there are no database schema changes since we don't have one


mike3985 (approved):
Seems reasonable to me. SRTM

PR #1670, opened 2026-09-09
[PP-7714] Pact v2 migration

geek4good:
Migrate pact verification tests to pact v2 ([Jira ticket](https://gov-uk.atlassian.net/browse/PP-7714))

Upgrade pact from v1.x to v2.x and rewrite the tests for the new Rust-core based API.

- include Pact tests in the CI run
- bump pact to ~> 2.0 and add pact-ffi (v2 verification runs on the Pact Rust core via FFI) and the rspec meta-gem (pact 2 no longer declares it as a runtime dependency)
- replace the v1 DSL in spec/service_consumers/pact_helper.rb with the v2 RSpec DSL in spec/pact/consumers/publishing_api_spec.rb
- move per-interaction setup into Pact.configuration global hooks so auth stubbing and database cleaning run for every interaction
- define the pact:verify task explicitly in the Rakefile (pact v2 no longer ships lib/pact/tasks.rb)


spec/pact/consumers/publishing_api_spec.rb
```diff
@@ -0,0 +1,46 @@
+require "rails_helper"
+require "pact/v2/rspec"
+
+RSpec.describe "Verify consumers for Content Store", :pact_v2 do
+  Pact::V2.configure do |config|
+    config.before_provider_state_setup do
+      DatabaseCleaner.strategy = :truncation
```
callumknights: I'd be interested to know exactly why the switch to `:truncation` was necessary. A quick search suggests the RUST verifier is just inherently more difficult with transactions. But if you can still recall, some detail in the commit message would be really useful for posterity.

Rakefile
```diff
@@ -4,6 +4,22 @@ rescue LoadError
...
+    task.pattern = "spec/pact/consumers/**/*_spec.rb"
+    task.rspec_opts = "--require rails_helper --tag pact_v2"
+  end
+
+  namespace :pact do
+    desc "Pact v2 verification"
+    task verify_v2: :pact_verify_v2
+  end
+rescue LoadError
+  # Pact isn't available in all environments
```
brucebolt, line 20: Nitpick: is it worth rewording this to `Rspec and Pact aren't available in all environments` (to cover the missing `rspec/core/rake_task` scenario we saw yesterday)?

brucebolt (approved):

